PR #4, opened 2018-12-06
Vent board fixes- fixed

AlexTheDragon8:
- Switched Q3 and Q4's sources and drains
- Connected R13 and R18 to +5VBAT (instead of +BATT)


amihaila:
Merged manually

PR #6, opened 2020-01-13
Gps 2

JaredWatson:
Continuing from the review on slack


zx100x100:
Please add a PDF of the Schematic to the GPS 2 Folder

amihaila:
> Please add a PDF of the Schematic to the GPS 2 Folder

I don't think we should be tracking kicad pdfs in github. Having pdfs fall out of sync with the real schematic (which will inevitably happen) isn't great. Keep your pdf posting needs to slack.

amihaila:
I don't have anything to add on top of Aaron's comments.

zx100x100:
> > Please add a PDF of the Schematic to the GPS 2 Folder
> 
> I don't think we should be tracking kicad pdfs in github. Having pdfs fall out of sync with the real schematic (which will inevitably happen) isn't great. Keep your pdf posting needs to slack.

Ok, I think most of canhw has a pdf in git so I assumed this was standard.

AidanHa:
>Why are you using 4800 baud hardware setup instead of 9600?

I think the previous board used 4800 so we wanted to stick with that. If you want, we can change it to 9600

>You're cutting dropout voltage pretty low on aux batt with a 3.7V battery, at 100mA you have 80mV typical LDO dropout and 280mV typical forward voltage, so there's only a 40mV of wiggle room. Not sure what else you can do though, short of using a higher voltage battery (5V USB backup battery?)

@AlexTheDragon8 You were responsible for this change, thoughts? I think having a 5V battery on standby might help.

JaredWatson:
I misspoke on the update. It's actually U1. Guess I should be more careful with that

zx100x100:
I think this may have been brought up before but is there a reason the GPS RX isn't connected to the PIC? I'm not sure how useful it would actually be but it seems like a very low opportunity cost and it would feel pretty bad if you end up needing it.

JaredWatson:
> I think this may have been brought up before but is there a reason the GPS RX isn't connected to the PIC? I'm not sure how useful it would actually be but it seems like a very low opportunity cost and it would feel pretty bad if you end up needing it.

It is, see the lines `GPS_RX_PIC_TX` and `GPS_TX_PIC_RX` on U2 and U1

zx100x100:
> > I think this may have been brought up before but is there a reason the GPS RX isn't connected to the PIC? I'm not sure how useful it would actually be but it seems like a very low opportunity cost and it would feel pretty bad if you end up needing it.
> 
> It is, see the lines `GPS_RX_PIC_TX` and `GPS_TX_PIC_RX` on U2 and U1

Oh shoot sorry, I've been looking at an old schematic

zx100x100:
A few notes on your initial component layout:
1. It looks like you are using a terminal block for your external battery, is there a specific reason for this? It takes up a huge amount of space and will make the board a lot more unruly than It should be, I would switch to like a 2 pin JST.
1.a) You have the terminal block on the bottom of the board, move it to the top if you need it because otherwise you will need really tall standoffs to actually mount the board.
1.b) C10 currently intersects the terminal block.

2.Right now a lot of your components are really close together, and it will be a pain in the ass if you need to replace a part. Try to spread them out a little and preferably orient them all in the same direction so they look a little more organized and are a little easier to install / replace. I think vent is a nice example of this:
image

It's ok if you need to make the board a little bigger, you have it squeezed into 1.5x1.75", I think as long as you keep it less than a 2" square size isn't too big a deal.

3. There are some things you need to keep in mind for your rf line from the antenna to the GPS. This article goes into a lot of detail, but essentially you will need the GPS on the same side of the board as the antenna connector, you will need to isolate it with ground planes, and you will need to use a specific trace width to get the correct trace impedance (I believe 120 mil)
https://www.electronicdesign.com/technologies/communications/article/21799702/adding-a-gps-chipset-to-your-next-design-is-easy

4. while it doesn't need to be perfect there are some guidelines about the placement of your crystal. mainly that it should be near the two pins it connects to on the pic, and the bypass capacitors should be beside it. section 2.5 of the PIC data sheet explains this in detail, and I think remote arming is the best example of how it should be implemented..
http://ww1.microchip.com/downloads/en/DeviceDoc/PIC18LF25-26K83-data-sheet-40001943B.pdf

akmorrison:
Things that Zach didn't cover:
You should add a board outline on the Edge.Cuts layer so that the fab knows exactly how big your board should be. To do so, click on the Edge.Cuts layer in the sidebar so the blue arrow thing shows up on the left of it, then the "Add Graphic Lines" button in the right side toolbar thing. Then draw a square around your board.
image
It should look like this when you're done
image

Other than that I agree with everything Zach said. As a clarification to his "they're too close together" point, try to think about where you need to position the soldering Iron in order to solder that pad, then see if the iron is on top of anything. Typically, you put the iron in parallel with a pad to solder it, so things like pin 1 of the programming header can't be soldered without melting the microcontroller:
image

zx100x100:
Looks Good!
I might put R6 in  line with R5 for aesthetic reasons also the LEDs D2 and D3 and their resistors are 
image

Also the LEDs D2 and D3 and their resistors are a little squished in there right now but I'm not sure where I'd put them so it's probably fine.

I would also orient the silkscreen labels all the same way , but that definitely doesn't need to be done before you route your traces.

Generally looks like you are good to do trace routing.

akmorrison:
Ok so I know that you've said that you're reworking the layout, but I'm gonna put down my thoughts on 2c74dce. I've already mentioned the "don't run an RF trace between the crystal and the microcontroller", and that's the only big thing that needs to be changed. However, I would like to write down some general guidelines on component spacing because I always complain that the spacing is wrong (not just on this board, all the boards).

For 0805 components, my general rule of thumb is that you want them lined up in a row so that the major axis of the component (the line from pad 1 to pad 2) of all of them is parallel, and that their minor axes are all on the same line. Like you've got with C1, C7, and C8. A couple more restrictions I'd add:
1. There should be enough space between them for the silkscreen component reference. This is for aesthetic purposes and to make sure you have enough room to get a soldering iron on one component without touching its neighbours.
image
C1 is just a touch too close to C8, and C7 is more too close to C1.
2. The spacing between components lined up in a row should be consistent. This is entirely for aesthetic reasons.
image
The spacing between these passives are, respectively, 115mil, 150mil, and 115mil.

The easiest way to solve both of these is to increase your grid size when placing them. With my FBE approach of "make the component reference fit between them", you want about 140 mils from the center of one component to its neighbor.

For components lined up on their major axis (or with a component next to another thing, so that the direction of soldering is facing it), you need more space. I don't have a rule of thumb for this, but there's a few places on the layout where you've got things too close.
- Pad 2 of R9 is too close to pad 2 of F1
- Pad 1 of R9 is too close to pad 8 of U1
- C3 is _just_ far enough away from pads 1,2,3 of U3. I wouldn't go any closer than that
- D2 is goo close to pads 3 and 4 of J1
- D4 is an ok distance from C4, but pad 2 of D4 is too close to C7

JaredWatson:
I've addressed all of the comments you have, and combed the [datasheet](https://antenova.com/wp-content/uploads/2015/11/GPS_Radionova_M10578-A2_PS-1.0.pdf) for anything about running a trace under it but I can't find anything. The only note is that it's solder temperature (page 17) should be 245C±5C

amihaila:
As mentioned for a few other boards, make sure to add silkscreen indicating which programming header pin is MCLR, and what voltage the PIC runs on like this:

image

zx100x100:
Looks good, two minor things:
1. Can you add the technical name of the board in addition to the fun name?
2. I would move your MCLR marking to make it clear which pin is MCLR (also right now it reads as J13V3 MCLR)

I could be missing a bunch of stuff but I think you are good to move on to gerber review.

PR #7, opened 2020-01-13
Recovery

zx100x100:
Please review initial PCB component layout

### DIFF
--- a/recovery/sym-lib-table
+++ b/recovery/sym-lib-table
@@ -1,3 +1,4 @@
 (sym_lib_table
   (lib (name recovery_CAN-rescue)(type Legacy)(uri ${KIPRJMOD}/recovery_CAN-rescue.lib)(options "")(descr ""))
+  (lib (name canhw)(type Legacy)(uri ${KIPRJMOD}/../lib/canhw.lib)(options "")(descr ""))
 )


akmorrison:
The board is really big and has a lot of unused space, you can probably shrink it down by a lot. Typically you're bounded by connectors more so than components, so my typical strategy is to put the connectors around the edge of the board in a squarish shape, as tightly packed as they can be, then layout the components in the middle. More concrete shrinking advice:, you can move a bunch of stuff to the back side. You can probably move D4/D5 to the bottom, same with U1, Q1,5,2, and most of the resistors.

akmorrison:
Other thing, a few of these components are gonna be impossible to solder in/replace once the connectors are in. You won't be able to get an iron on the the leads of D4/D5, or the fat pads on Q3 and Q4, since they're blocked by the connectors next to them.

akmorrison:
I'm willing to bet you can get the board down to 2.75" square, or maybe 2.75 high by 3 wide

amihaila:
(Things Aaron didn't mention/things he did mention that I wanted to elaborate on)

You can improve the way that higher power section is laid out:
image
Right now it's very spread out, with a bunch of unrelated circuitry interspersed within it. As a result, you're forced to route quite a few thick traces around the edge of the board. I would try to physically group all of that circuitry and keep that higher power section self-contained.

The way I typically do this is I take the relevant connectors, look at the rats nest between them (without actually drawing traces) and arrange them so that traces have to travel the minimum necessary distance between connectors. Then I start bringing in related large components. This will likely change the shape of your board. One possible solution might be to align J5, J10, and J6 along the bottom edge of the board, and to align J7 and J9 along the right. J8 is out of place where it is right now - none of those long traces connect to it and you're forced to route around it.

Putting your jumpers (eg. JP7 and JP9) in between your connectors also takes up unnecessary space along the edge of the board. Moving these further in, or to the other side of the board, would help you save that space and make your board smaller.

**Misc. other things:**
* How much current draw are you expecting? Are those 40 mil traces really necessary? (nitpick: do your design in imperial so those traces are actually 40 mil and not 39.37008 mil. Most PCB design I've run into is in imperial and we follow that convention.)
* Try to aim for some visual flow in your board layout. Keep components aligned and in the same orientation when possible. Similarly to a nicely laid out schematic, a nicely laid out PCB is easier to deal with, especially when you're debugging it.
Your PCB doesn't need to _look_ like your schematic, but it is useful to be able to easily translate between the schematic and PCB layout. It will also help you place your components in a less haphazard and more compact way. A nice example is one of Adam's boards from last year:
image

zx100x100:
I'll change the schematic to address these. Two things:
- I'm expecting current draw of up to 2 amps (although only for a second) as far as I understand it the rule of thumb is 40mil per amp, so 40mil is actually half the recommended width. (although since it is only for a second I think it's fine.) If you think I can get away with thinner then great!

- I'd like to keep the connectors on the sides they where already on. This layout of connectors is something I discussed with roman as one that should be fairly logical and clean when wiring up the recovery electronics. It looks like I should be able to hit Aaron's 2.75" square target without moving them. That beign said if it is a big deal for some reason I can move them.

pauladam316:
Mostly just echoing what alex and Aaron said, but a few more things to note:

Try to group components by submodule. Eg on thonk I have an altimeter that needs a i2c level shifter, so I put those components along with their filter caps and pullup resistors into a tight grouping and found optimal connections within the module, that way you can move it around as a unit easier and you don't have traces going everywhere

Make sure your part labels aren't in the footprints themselves. Also try to orient them all one way, it looks nicer

You can hide the "REF**" designators on the mounting holes and I would recommend doing so

I'd also recommend pulling the connector CAD from the harwin website and adding it to the footprint. It will give you a better spacial sense of how this will fit into the rocket when looking at 3d view and might reveal connection issues

zx100x100:
Can I please get a review of this new component layout?

zx100x100:
Working on addressing these, the only thing is the connectors:
It will be a bit of a pain if everything is attached to one connector during assembly and wiring (not impossible but a bit annoying), and it would take a fair bit of work to change the board for this. 
For now I am going to assume we can get the connectors (even if we can't I can do testing and codding on a board without them) and then I will find new connectors and re-design the board if we cannot get them.
If this doesn't seem reasonable let me know.

zx100x100:
> Few things
> -component silkscreens aren't properly laid out. some are horizontal, some are vertical, some overlap and some are off the board 
I assume the off the board ones where J2, and J3 (moved now), Which ones overlap? also some are vertical and some are horizontal since they align with the part, although they are all either above or left. Is this not correct? do they all need to be on the same side?

> -Oscillator circuits require a keep-out below them to have a consistent ground plane. Your caps are currently above a connector. I'd recommend moving them
Going to move the caps but what do you mean by a keep out for a consistent ground plane?

pauladam316:
IIRC there was a diode and a resistor that overlap. Its up to you how you want to orient the silkscreen but imo it looks much cleaner with them all oriented the same way.

by consistent ground plane I mean that the ground pour below the oscillator circuit shouldn't be broken up by traces or pads. Should be one continuous piece

zx100x100:
> This looks a lot better. I give it a conditional approval with the caveat that:
> You check that the footprints for U6 and J2 (similarly for U7 and J3) are accurate. These components are very close together and if your footprints are off you're going to have a bad time. This has happened to me and Aaron before and we ended up having to file parts of a connector off.
> 
> Aaron already brought this up but I didn't notice you commenting on it, so I'm bringing it up again as a potential trouble spot.

Sorry, forgot to mention earlier, but yes I have checked and the spacing isn't a problem.
> 
> Nitpicks that you don't need to change, but you asked me to comment on:
> 
> * You misspelled "Drogue" on the silkscreen.
Fixed
> * Don't forget to fill out the date on your silkscreen.
Thanks for the reminder
> * Typically you want to constrain the bends in your traces to 45 degrees. You have a lot of traces making 90 or greater bends after vias. Reasons why you shouldn't do this range from signal integrity (not an issue on this board - you're not running any signals in the MHz range) and manufacturing (not really an issue in this day and age anymore). Examples:
>   image
>   image

This is good to know, thanks! (not going to bother fixing it for this board)

> * This trace is a bit wonky. 
I tidied it up a tiny bit, although I'm not going to try and redo a bunch of other traces to make this one super clean.
> It switches layers a bunch of times and isn't very organized. It also doesn't need to be anywhere near that fat - that buzzer only draws 30 mA.
The portion of the trace that goes to the buzzer is only 8 mils, the part that connects to pyros is 15 mils, and the parts that power the altimeters is 30 mils.
>   image

AidanHa:
REVIEW:

Maybe I'm out of the loop since this is my first time reviewing the board but why is there traces in the No connect zone?

image

I also remember getting roast last year for having my caps (in your case C1 and C2) way to close to the crystal, but I can't seem to find the post on slack...

Otherwise, LGTM :)

zx100x100:
> REVIEW:
> 
> Maybe I'm out of the loop since this is my first time reviewing the board but why is there traces in the No connect zone?

Not sure exactly which traces you are talking about and what you are considering the No connect zone, but essentially I did my best to keep the traces away from the crystal, but it's not perfect. As is, this is probably the most correct isolation currently present on any of our boards.
> 
> image
> 
> I also remember getting roast last year for having my caps (in your case C1 and C2) way to close to the crystal, but I can't seem to find the post on slack...

Yeah, I think this is kinda a trade-off in that if I need to replace those caps it will be a pain, but this was approximately the distance they where away from the crystal on the data sheet.
> 
> Otherwise, LGTM :)

PR #8, opened 2020-01-17
Live Telemetry

dwkletke:
Schematic review continuing from slack.


zx100x100:
3 Comments:
You are using 330 ohm resistors despite powering the LEDs from 3.3V, not 5V, you should be using a value of 130-140 ohms

Looks like the frequency for the radio is programmable which is good, means there is no issue with setting it to a frequency assigned by ESRA. (please do still ask about the Xbee in the BigRedBee)

Not sure if it would be useful, but if you think there is any utility in the Pic being able to reset the Radio you might want to add a solder jumper that lets you connect reset to IO on the pic.

akmorrison:
I agree with zach, the reset input should be driven by the microcontroller, and shouldn't be connected to a button. Buttons have a possibility of accidentally closing due to launch acceleration/vibrations, and it would suck for our radio to reset while in flight

akmorrison:
Otherwise, I don't see anything wrong, it looks good!

akmorrison:
Change looks good, but you should change the value field of the microcontroller on the receiver/client side to be the low power version, right now on my machine it still shows the non-L version

image

akmorrison:
Similar points for the receiver board: Wrong coax footprint (should be on egdge of board), move transceiver to keep your RF trace straight, you could make the board shorter if you don't put the LED resistors on top of the LEDs, they can go beside them.

zx100x100:
Looks good, Just a couple small things:
On receiver:
1. I would rotate C1 and C2 180 degrees so that the 3.3V pins are in line with the one on the radio
image

2. Is there a reason you are using a male programming header instead a female one like all the other boards?

3.  You cold make the board a bit smaller without making it much harder to assemble if you move the LEDs, programming header, and Coax connector over. (it would make it slightly harder to replace  C1 and C3 but no too bad, and filter caps rarely need to be replaced.)
image

4. I might move R3 and R4 down a little to give yourself plenty of room to solder (it's not bad how it is right now, but you might as well since you have the space)

I think it looks good and once you address these you can start tracing.
I'll post my review of receiver in a few minutes.

amihaila:
Just a note to everyone: In general I don't think that layout nitpicks such as "X small component can be rotated Y amount" need to be addressed in layout review. It's expected that the layout will be tweaked when traces are added.

amihaila:
Don't forget to do this later: on the programming connector, indicate which pin is MCLR and what voltage the microcontroller runs on. Both the transmitter and receiver are missing this. e.g.

image

JaredWatson:
1. Receiver: A lot of your traces are daisy chained:
image
 this is bad if the pad lifts you lose everything downstream. Try to do something like this
image
instead (there's also the GND net there that should be fixed). You did a pretty good job on the transmitter, although some parts are still connected in series.

2. Receiver: This MCLR trace goes through 3 vias because the 3V3 traces were stopping it, with a pour this is resolved and it only takes one.
image

3. Transmitter: Datasheet recommends that the oscillator, MCU and bypass capacitors are on the same side of the board, shouldn't be hard to flip the side and rotate Y1 around. Don't forget the isolated ground pour.
image

4. I know you said you would add the pours on your next commit but you ended up routing power and ground anyways. When you're making the pour I recommend GND on the bottom and 3V3 on the top for receiver and for transmitter GND bottom and either 3V3/5V/GND on top, whatever makes your life easier. All of those are suggestions, not a hard rule.

5. You can save more tracing space if you use separate vias on all your tracks opposite their pour, instead of connecting them together through a single via. They also provide thermal, relief instead of sending all your power through one via, it gets distributed.

Overall looks good so far, give the pours a go, cut out the daisy chains and see how the routing feels.

akmorrison:
All that is on transmitter. on receiver maybe move C2 R3 R5 and C4 away from U1, move C6 and C5 away from U3, but other than that it's fine

PR #9, opened 2020-02-04
Fixed Footprint & Library paths

LanaTomlin:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/9)


JaredWatson:
They should probably be added to the gitignore but I noticed a lot of .bck and .lib files were also changed. Unless these were intentional you can elect to omit them from your commit by adding just the lib tables to the commit. Using `git add <file>` for each of your tables instead of `git add -A` will solve this. 

I also noticed things like `picdev/picdev/picdev.sch` were being changed (see lines 3, 45, 225 etc.) where 225 had a CAN Breakout changed to a Conn_01x02_Female, and this wasn't reflected in the commit message.

LanaTomlin:
I tried to remove the unnecessary changes caused by me re-mapping symbols that didn't show up on my computer, but now it says "merge branch" I am scared I or may not have broken something in the process. I apologize.

jacobdeery:
What's the status of this PR? @LanaTomlin

PR #11, opened 2020-02-16
Logger mini layout

amihaila:
Main changes with respect to the previous logger board:
- The second (unused) SD card slot is gone
- The CAN connector is replaced with a through hole version
- Both sides of the board are now used so it's smol


akmorrison:
Overall good, I've got some comments that kinda matter, and some that don't really matter and are totally for style points. I'd prefer if you fixed the former, I don't really care about the latter

Things that kinda matter:
- Mounting hole dimensions are off, the eco1.user says .84x1.4, if I measure center to center I get .849x1.397. My kicad shows TL hole at (5.031,2.683), BR hole at (5.88,4.08).
- J4 silkscreen reference is on top of CAN connector footprint
- It looks like it's gonna be hard to solder/desolder pad 1 of D2 once the connector's in place, you could move the LEDs down a bit
- Current sense path from pad 1 of R5 to pad 4 of U4 is a little borked, the path to the current sense amp should be separate from the main power path, there's a little bit of shared wire:
image

Style point things (very petty, can be safely ignored):
- If you rotate C10 180 degrees it'd be a slightly shorter path to the component it's bypassing
- If you move C10 35 mils to the left, it wouldn't be below the voltage regulator. It wouldn't be in the path of the soldering iron then
- You have room to move C2 & R1 away from the programming header, would make it slightly easier to solder.

amihaila:
@zx100x100 in response to your comments:
1. I moved C9 and C10 to be in line
2. I put the caps near the MCU cause it doesn't matter that much and I wanted to leave more clearance near the MCU pins (instead of putting a tall crystal there)
3. I'm not sure what your issue with C2 is. It bypasses pin 28 like so:
image
The via you're pointing at in your drawing is the 3v3 connection to the PIC programming header.

zx100x100:
1. rad!
2. Totally fair
3. Sorry, I'm an idiot and got confused between the via for the filter cap, and the 3.3V line on the programming header.

AidanHa:
LGTM, I couldn't find any glaring issues other than the board not having a date/author.

amihaila:
@akmorrison 
I fixed the mounting holes and the current sense amp traces. I'm ignoring your style nitpicks.
I also changed the name to RocketCAN LOGGER cause idk.

Mounting holes should be accurate now. The kicad grid was doing weird shit last time.
Screen Shot 2020-02-17 at 2 45 27 PM
Screen Shot 2020-02-17 at 2 45 45 PM

amihaila:
Ready for fab review

akmorrison:
How do you feel about including zip files in fab review? It was convenient with minisensor to have the zip file already on the filesystem when I went to order more of them

amihaila:
I figured I'd commit them when there aren't any more changes to make. No point committing the zip otherwise. All it does is invite the possibility of the gerbers and zip falling out of sync.

akmorrison:
Otherwise, the gerbers look good to me. I vote to call this board ready to order

zx100x100:
Gerbers look good, looks ready to order

PR #12, opened 2020-02-17
Vent 2

MattiGen:


JaredWatson:
Here's the footprint I used 
image
and here's the datasheet's pinout

JaredWatson:
Those buttons are too close together, sorry

JaredWatson:
As for your 12V line question, I really didn't have any reason in mind. I think I was trying to shrink it a little by moving Q1 to the back and thought that it would be better in terms of the PIC pins to place it there, but after thinking it over I don't know how much it saved me. If there's a real reason to shorten it I can move the connector back but otherwise I feel like leaving it where it is, simply because I don't see an effort/reward balance. Again, if there's a serious reason to make the 12V trace as short as possible I will move it.

zx100x100:
Looks good, a couple comments:
1. Can you add the part number of the mosfet to the schematic 
2. You are doing something funky with C3, the filter cap for U3, where it goes through 2 vias before getting to the +5V pin
image
I would do something like this:
image

3. your isolation for the crystal doesn't really follow the datasheet, but this is real nitpicky, and what you have will totally work.
image

zx100x100:
There is also a bit of a long trace for your filter cap C2 for the pic:
image
This can be easily fixed:
image

zx100x100:
Another long trace on C1 to U1:
image

I think if you run CANH around the cap you should be good. Also to help avoid this type of thing where the fill creates a long trace I like routing the power traces between my filter-caps, and their components before doing pours.

zx100x100:
Sorry, two last things:
1. I might move J1 over a bit, it's kinda cramped in a corner and it doesn't need to be.
2. I relay don't think it's worth fixing, but just for these parts will be a pain to replace once J2 is soldered on.
image

zx100x100:
Sorry one last one, you have a via on a pad:
image

akmorrison:
WRT Alex saying it's easy to fix: You easily have enough room on the back of the board to fit those 4 components if you move R5, 6, 7, 8, and 9 to the right a ways. If they're on the back they'll be easy to replace even once the connector is in place

zx100x100:
Looks really good! :_heart_eyes: only one little thing:
1. The weirdly long path for filter-caps has struck again:
image
Easy fix:
image

PR #13, opened 2020-02-18
Fill sensing

zx100x100:


zx100x100:
In response to Aaron's comments:
1. it's a molex connector, I couldn't find any JSTs I liked 
https://www.digikey.com/product-detail/en/molex/1053141116/WM16913-ND/7693194

2. It doesn't seem bad
image

3. oh, good point. I hadn't really figured out which layer these where supposed to live on

4. Yes, this cable harness will be a pain in the but. Maybe Aaron will decide he wants a few less but I'm guessing not. The idea here is it's not hard to support up to 14, and if we only want like 8 then we can only wire up 8.

5. Fixed, thanks

6. yeah probably, but that just seems a little excessive. given how small the board is I don't mind if it's a tiny bit bigger if that makes it easier to work with. also I like having the can connector on the opposite side of the board as the Molex, as it will make wiring this up on the rocket easier.
image
That being said I think you are right, and I am going to try and shrink the board a bit more.

zx100x100:
In response to Jared's comments:
1. As stated the routing seems fine

2. oh lol, nice catch.

3. I don't think it really matters, but that does look prettier so I'll fix that

amihaila:
If you fix these two things, you'll have my approval.

JaredWatson:
- Just wondering, are these traces supposed to be run through J3? Does this matter?
image

- I'm not sure if this matters as much with silkscreen but the MCLR marking is in R4's keepout
image

- Is the plan to plug the hall sensors power from the bottom? Currently the silk is on the bottom and might be hard to read while assembling (If the plan was the bottom you can ignore this)

Other than that it looks great! When these are fixed I don't think I have any other comments.

zx100x100:
I'll move the MCLR over a bit,
 Traces through J3 should be fine, if we lose the pad, losing the pull up doesn't really matter.
The sensors attach to a right angle on the top of the board, you won't be able to read the silkscreen if it is on the top

PR #20, opened 2020-05-01
Switch minisensor to run off 3v3, not 1v8

akmorrison:
We weren't able to program the microcontroller on minisensor when it was
running on 1v8 VDD, so we swapped the voltage regulator for a 3v3
regulator and it all worked fine. The original reason for 1v8 was that
there was a digital sensor we wanted to use that couldn't run on 3 volts
(the ICM-20948). This sensor got removed in the move from macrosensor to
minisensor, so there's no reason not to use 3v3 on minisensor. The three
digital sensors on the latter all work at 3v3, datasheets below:

- https://www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=MS5607-02BA03&DocType=Data+Sheet&DocLang=English
- https://www.st.com/content/ccc/resource/technical/document/datasheet/group0/38/96/79/af/ab/62/42/c3/DM00229854/files/DM00229854.pdf/jcr:content/translations/en.DM00229854.pdf
- https://www.st.com/content/ccc/resource/technical/document/datasheet/74/c4/19/54/62/c5/46/13/DM00177685.pdf/files/DM00177685.pdf/jcr:content/translations/en.DM00177685.pdf


QuantumManiac:
One more thing to bring up actually, but not much of an issue. It seems that the `Team_Logo` footprint library is set to an absolute path rather than a relative path like it should be:
image

PR #21, opened 2020-05-04
Live telemetry transceiver breakout

QuantumManiac:
Okay, here's my first pass at the schematic for the live telemetry radio module breakout board.

In addition to the usual review, please let me know of any ways for me to improve the layout of the schematic if you can - this is my first one after all, and I'd image that there's plenty of room for improvement.

Relevant datasheets:
[TXB0106 Bidirectional Level Shifter](http://www.ti.com/lit/ds/symlink/txb0106.pdf?ts=1588599768745)
[LP5907 3.3V Voltage Regulator](http://www.ti.com/lit/ds/symlink/lp5907.pdf?ts=1588599843795)
[RFM95W Transceiver Module](https://www.rfsolutions.co.uk/downloads/1463993415RFM95_96_97_98W.pdf)

---
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/21)


zx100x100:
Looks good to me, maybe one or two ways you could make the schematic more pleasing, but no real changes.

image
I would adjust this so the wires don't intersect, same on U2.

image
I think you could line things up better to make this a smidge prettier

dwkletke:
I agree with Zach. The only things are aesthetics.
image
I think this could be cleaned up on both U2 and U3 by aligning the capacitor with the GND or something of that nature.

QuantumManiac:
Okay, did a bit of cleanup. The caps for the level shifter look a lot better but let me know if this is any better for the reset switch:
image


Additionally, I noticed that I've been using 3V3 and 3.3V sources interchangeably so I fixed that.

dwkletke:
Overall this layout and routing technique looks good. There is just a few things I would like to add.
image
This big rf trace should be straight. May need to make the board a big bigger to accommodate this.
image
I also feel like these capacitors and resistors are too close together that it might be a pain to solder/replace if something happens. Additionally, it should be noted that a lot of these capacitors are bypass/decoupling capacitors. Due to this nature, they should be as close as possible/relatively close to the pins they are supposed to decouple (you should reference your schematic here).
image
There is a 90 degree trace here, biggest angle should 45 degrees.

dwkletke:
There is one thing I forgot to mention:
image
The caps/resistor VCC and GND traces shown here are daisy chained which could be an issue if one of the components fails. They should be individually connected to the GND net, ex:
image
Obviously without covering the component labels, but your routing may change if the caps move.

QuantumManiac:
Okay, permitting acute angles for traces definitely does clean up a lot of the junctions. I wasn't too sure whether to use them or not due to the notion of them causing acid traps during PCB etching and thus not being considered best practice by some.

JaredWatson:
Unless you're making them in your garage it's fine. The aesthetic is more important anyways!

I also noticed you seemed to be using the footprint I made for the SMA connector. Don't it's BAD and HARD to solder. Make your life easier and make the pad size bigger. I haven't cloned the repo so I'm not sure if it's the same or similar but to my eye it looked to be the same. Message me on slack if you have any questions about it and we can post a summary to electrical if/when we "update" this footprint everyone has unfortunately copied.

zx100x100:
I have a few comments:
1. I would switch the footprint on the additional IO pins to a normal header, not a 90 degree header. As is we don't intend to actually use those pins unless we want to do something weird, and the 90 degree header takes up a lot of space.
I would also move it to an edge of the board away from the antenna as it is likely the SMA connector will go through a hole in the enclosure and anything on that side of the board will be hard to get at.

2. I don't know if is worth it, but I personally like it if the SMA connector is down the middle of the board, you could probably do this in a kinda cute way by rotating the module 45 degrees.
This is probably a fair bit of rerouting, so if you think I'm crazy then feel free to ignore me.

3. ignore me about the 90 degree JST connector, If we decide we want one later it can still be attached, it will just hang off the side a little.

4. Try to make sure the reset button is as accessible as possible, try to put it near the edge of the board away from any tall parts. (it isn't bad right now, i just think it could be a little better.

5. In general I think you can probably get the board a little smaller, there is a fair amount of unused space on the board and as we are likely to use it in ST_PATS the smaller the better.

image

Overall for an initial layout it looks great.

QuantumManiac:
Okay. Header and reset pins have been moved to more appropriate locations. From this, the board size has been reduced to 1.5"x1.5" as seen below:
image

I did attempt to center the SMA connector by rotating the module 45 degrees but spacing became quite tight and routing became a bit of a nightmare so I decided not to proceed. Would've looked cool though.

QuantumManiac:
Two things to bring up:

- This is what I came up with for RF isolation:
image
 I wasn't too sure as to what the methodology was to this so I pretty much took a look at the other boards and did something similar.

- Due to C3 and C4 blocking the area beside the JST connector, I just wrote a list of the pins instead of having the labeling line up:
image
 Would this be okay in this case?

zx100x100:
> Two things to bring up:
> 
> * This is what I came up with for RF isolation:
>   image
>   I wasn't too sure as to what the methodology was to this so I pretty much took a look at the other boards and did something similar.
This looks pretty good, the vias near the bottom probably don't do much, but it's ok. 
> * Due to C3 and C4 blocking the area beside the JST connector, I just wrote a list of the pins instead of having the labeling line up:
Sorry I probably should have been clear, I think you should have the 5V regulator, and it's filtering caps, on the bottom, that will let you line up the labels, and make the filtering caps easier to replace. 
>   image
>   Would this be okay in this case?

Also don't forget to hit b to refill your copper pours before committing, I think there is a couple spots in this commit where the copper isn't filled in right.

zx100x100:
this one mounting hole is like 5 mils out of alignment.
image

you probably could get the bypass caps a little closer to their respective ICs, but it isn't a big deal.

I would include Gerbers for review in your next commit.

QuantumManiac:
Requested changes have been made. @zx100x100 could you possibly give me some insight into where I can place the decoupling caps to get them closer to their respective ICs? I'm taking a look at the PCB and in my mind it seems to be quite difficult to get the traces too much shorter without getting components really close or drastically changing the routing.

zx100x100:
for the voltage regulator You can put the decoupling caps on either side.
you could also shorten the trace for C2, the rest look good.
image

QuantumManiac:
> The only thing I would add is that you need to hide the REF** labels for the mounting holes since they are still visible on the silk screen.
> image
> This is done by right-clicking on the label -> Properties -> deselect "Visible" field.

Good catch. I suspect that it slipped through while I was changing the size of the mounting holes a few commits ago. Fix has been made.

lwbantoto:
Schematic notes:
- Never have a 4 wire junctions
image
- This is how I typically avoid 4 wire junctions
image

- I would add a decoupling capacitor for U4. It can't hurt.
- I would add an RC debouncing circuit for the reset pin for U4 (10ms debounce time is fine)
- On my schematics, I like to have the capacitor rated voltage specified and the resistor package specified. This lets the reader know if the capacitor is properly derated for the voltage it "sees" and it lets the reader know the power that each resistor can withstand.
- For SPI, consider using a pulldown resistor on-board for NSS.

lwbantoto:
Add some test points :)

akmorrison:
>     * Never have a 4 wire junctions

This is a kicad vs eagle thing, but those aren't actually 4 wire junctions. When you have non-connected crossing wires, kicad draws them as straight lines (crosses). When they are connected, kicad puts a solder dot on them (like the junction just beside the VCCA pin)

I'm not sure if kicad even lets you draw the little jumps in the wires like you've got in the second screenshot

JaredWatson (commented):
So far looks great! A few small comments that could be addressed (although I see you're on gerbers and they're not critical)

- The clearance on your pours is 20 mils, it _could_ be less (mine was 6) but that just makes it look a little nicer and might fill some gaps like here
image
- This section around U1 looks a little cramped and you could easily make space for it without messing up the routing by just pushing them a little farther apart and make your life easier
image

That's really all I have to say. I know I'm reviewing this (a little) late but I wanna give massive props to the routing on this board, it looks amazing!

QuantumManiac:
Okay, we should be quite close to the final product at this point. I have addressed the various review points that everyone has given me, and this has resulted in quite a few things changing. Here's a run-down of the major changes:

- Pour clearance reduced significantly
- Un-cramped area around U1
- Added a power indicator LED
- Increased track width to 10 mils
- Specified minimum capacitor ratings in the schematic
- Specified max current for LDO in schematic
- Added a pull-up resistor for the NSS line
- Added test-points for GND and 3V3

Additionally, the gerbers have once again been updated.

zx100x100 (changes requested):
Everything looks good, except for a couple tiny little things

changes:
LED is currently 0603, we keep 1206 LEDs in stock so plz change.

also U1 is currently not using the hand soldering variant of the footprint, this isn't a huge deal, but it will be easier if you fix this

comments:
You picked a kinda funky test point shape, a like 1mm circle is most common, but that isn't an issue. It actually means it looks super easy to solder on 3.3V and GND wires, although no idea why we would want to.

also the current limiting resistor you picked is very small, this is fine as the white LED we keep around has a voltage drop of 3.3V, but I would leave a comment saying this is for a 3.3V drop as it would fry a like 2V LED if you installed that instead.

overall it's a really nice layout, I'm excited to get it ordered

zx100x100 (approved):
Awesome, everything looks good to me!

dwkletke (changes requested):
The only thing I see left is U1's silkscreen is currenting under a capacitor.
image
This should probably be moved under the part like this:
image
Other than that everything looks good to me!

dwkletke (approved):
Looks good to me!

JaredWatson (approved):
LGTM! Very happy with the aesthetics of the board and everything seems to be in order 👍

zx100x100:
Jacob has done some wizardry to get this merged, closing this PR and deleting branch.

Issue #22, opened 2020-05-10
RF footprint needs to be improved


JaredWatson:
Currently the RF footprint used on the GPS board and the Live Telemetry board had extremely small pads that make soldering it a pain in the butt. In the future the amount of soldering space should probably be expanded (but not enough to shrink the space of the RF trace, it's already smaller than 120 mils and we don't want to lose that sweet sweet RF integrity).

QuantumManiac:
I did a bit of testing, and it seems like it's pretty simple to have different sized pads on different layers.

In the _Pad Properties_ window, there's a Copper setting which allows you to change the layers that the copper and other layers are on for the pad:
![image](https://user-images.githubusercontent.com/8643561/81575435-07d3d000-9375-11ea-86b3-967c392c2a10.png)

My solution is to overlap two pads at each through-hole, with one pad having F.Cu and F.Mask and the other B.Cu and B.Mask 

Here's a view of a modified SMA connector in the footprint editor modified using the above method:
![image](https://user-images.githubusercontent.com/8643561/81575647-423d6d00-9375-11ea-88d5-d5957f690a35.png)
Notice that while the front pads are 2.2 mm, the back ones are much larger, at 3 mm 

3D Views - front and back comparison:
![image](https://user-images.githubusercontent.com/8643561/81576405-37370c80-9376-11ea-96d1-b686675f7087.png)

If I got some measurements for what the pads should be on each side, I would be happy to modify the footprint.

JaredWatson:
Looks good, I'll leave a more formal review on the actual PR. This is definitely the footprint people should be using for SMA stuff in the future.

PR #23, opened 2020-05-12
Increased size of pads on canhw RF SMA connector

QuantumManiac:
I created a footprint in the canhw_footprints library to address #22 in which the very small pads on the connector make it difficult to solder it onto the board.

This has been addressed by increasing the size of the pads on the back of the board to 3 mm (from 2.2). Let me know if this should be any bigger or smaller. The top pads have been left unchanged to keep the same clearance for the RF trace. Is there a possibility that we can further shrink these pads or get rid of them altogether for even more clearance?

To ensure compatibility, this footprint is a separate file from the original, named `RF_connector_SMA_3-1478978-1_BIG_PADS.kicad_mod`. I'd imagine that replacing the footprint outright would lead to issues in the boards that use the old one.

Some Pictures:
In the Footprint Editor:
image
Note how two pads are overlapped, with one being on F.Cu and the other on B.Cu

3D Views - front and back comparison:
image


QuantumManiac:
So, I did a bit of testing and it seems that in order to be able to fit the trace at 120 mils, we would have to get rid of the pads on that side entirely. This is based off of the default KiCAD design rules.

From that, I've got two possible options in mind:

- Get rid of the front pads conflicting with the trace entirely and only have the other two on the other side - this will give us the whole 120 mils of trace but I'm not sure what effect removing the pads entirely would have:
image


- Have the conflicting pads at the smallest possible reasonable diameter of 1.8 (drill is 1.6) and have the trace be a bit smaller at 112 mils:
image

Additionally, it seems that at 1.7 mm the pad is pretty much indistinguishable from no pad at all.

PR #25, opened 2020-06-26
Create V2 of remote arming :partyparrot:

zx100x100:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/25)


lwbantoto:
Please change diode symbol to Schottky symbol.
image
Typo
image
image
image
Avoid having 4 wire connections/crosses
image
Add a comment that says the crystal's frequency
image
Avoid overlapping text 
image
image
To reduce number of BOM line items, standardize pullup resistor value
image
Mag switches are missing their manufacturer part number
image

dwkletke (commented):
Overall, Remote Arming V2 looks beautiful! :_heart_eyes: 

A lot of my comments are style-related on the schematic and with the routing.

These symbol references/values are not the same orientation as the rest of the schematic:
image
image
image
image

This symbol value goes over a wire.
image

For the routing/layout:
I'd recommend moving this part reference underneath/somewhere else since there are 2 vias going through the silk screen:
image

Not sure what your opinion is, but I think it might be beneficial for the mag. switch symbol reference silk screens to be moved to be scene easily when they are mounted on the board:
image
image

The Q5 silk screen label looks like it is going to be covered by the connector:
image

Another nitpick: the C3 silkscreen is not the same orientation as the majority of components on the front side. However I am not sure if this was because of the abundance of vias in the surrounding area:
image

These bypass silkscreens look squished. Not sure if that was intended due to space:
image

Additionally the B of this bypass silk screen is cut off by the via:
image

Just checking: did you determine that this was enough clearance for the screw terminals and nearby traces? Just using this as an example:
image

Update date and version?
image

The J8 silk screen is off the board:
image

Nit: this trace seems really close to the R5 pads:
image

This trace feels really close to R6 pads as well:
image

What is our official stance on 90 degree angles for traces? Did we decide that for our non-RF applications, it's okay to have them?
image
image
image
image

Nit: U4 silk screen in via
image

Nit: more silk screens in vias
image
image
image
image

Is there a reason to why this trace isn't the same size as the rest of the net?
image

Same question as above for these:
image
image
image

Nit: this trace seems a little close to F1 pad1
image

zx100x100:
Thanks for the feedback Dawson! I've addressed most of it (I didn't fix all the silkscreen over vias where the silkscreen didn't go over the hole itself, just the coper, or any where the fix would have resulted in weird positioning or routing.)

As for the different width nets, I only used power traces when needed, if there is a thin trace coming off a wider one it is usually for voltage monitoring or something.

in terms of 90 degree traces they are fine for low speed stuff (non RF), but they generally don't look as pretty. I only used them where I thought a 45 degree trace would look more awkward.

I have now also generated gerbers for review

QuantumManiac (changes requested):
Super cool layout, very fun to just stare at for a while. I couldn't find anything off with the board to complain about, Dawson seems to have covered pretty much everything. 

I did notice that some project-specific footprint libraries were direct-pathed to your computer though. Do change them up to relative paths.

image

AidanHa:
Just finished my review, and had one comment: Could kindly explain the ground poor for the board. Idk if it's my Kicad or something but I see alot of GND vias, but no copper poor to connect them. I'm definitely missing something obvious because DRC check produces no errors. Maybe it's because I havent done a board review in a while.

P.S. the board layout is a fucking beauty.

dwkletke:
> Just finished my review, and had one comment: Could kindly explain the ground poor for the board. Idk if it's my Kicad or something but I see alot of GND vias, but no copper poor to connect them. I'm definitely missing something obvious because DRC check produces no errors. Maybe it's because I havent done a board review in a while.
> 
> P.S. the board layout is a fucking beauty.

I don't think you had the ground layer displayed. I'm not sure if you know but Zach recently made this a 4 layer board. The ground and power layers are not on the front and back layers, so that's probably why you don't see the ground vias connected to anything.

zx100x100:
done!

PR #34, opened 2020-12-29
Recover files for Live Telemetry Breakout

QuantumManiac:
When merging the original PR for the Live Telemetry Breakout, it seems that the files for the board rolled back a bit, resulting in an outdated iteration existing on master. In this PR, I just copied over the files from the commit right before it was merged, which contains the latest iteration of the board that was ordered a few months ago. 

This board was already fully reviewed prior to merge, but it can be looked over again if needed.

---
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/34)


zx100x100 (approved):
Opened up design files and they look good, how I remember them. I can see nothing else got tampered with so should be good to merge

LanaTomlin (approved):
Looks pretty

JaredWatson (approved):
Looks good! Everything seems to be in order here. Sorry for such a late response chamath <3

PR #37, opened 2021-05-28
PiCAM Board

ZTeper:
Please roast my schematic for the PiCAM board :)


BluCodeGH (changes requested):
Looking good, some small comments:
1. I've seen 200mR pretty commonly as the current sensing resistor, is there a reason you chose 300?
2. Add polyfuse value.
3. Consider connecting these two +5V lines that are right next to each other?
image
4. Remote arming laid out the MCP2562 really well, I'd suggest copying it:
image
5. Rotate your programming connector 180 degrees to match convention.
6. I haven't seen that MCLR resistor-capacitor combo before, mind explaining it?
7. Convention is CAN_TX on RC0, CAN_RX on RC1.
8. I think C8 would look cleaner next to Q2.
9. C6 is placed a bit awkwardly, move it slightly right?
10. Maybe give C7 its own GND and connect the left side grounds of J3 like you do the right side? You can pull 3V3 from pin 1 of the header to avoid a crossing too.
11. Remove "TEMP LBL"?

ZTeper:
The resistors and capacitors on the ~MCLR pin are wired according to the PIC18F's datasheet recommendations.

BluCodeGH (approved):
Cool LGTM. I'll just note the rotated text on C1, R4, D1, C8 for if you make another revision.

zx100x100 (changes requested):
Review time :partyparrot:

1. Really nitpicky sorry, but maybe remove _brd from the folder name? it's kinda redundant, and doesn't really fit with the naming for the rest of the boards.
2. RocketPiV1.pro doesn't point to anything, plz delete (along with RocketPiV1-cache.lib) Also kinda nitpicky but maybe consider renaming RocketPi to picam to match everything else?
3. we usually use 27pF caps on the oscillator, and we always use 12Mhz since that's what we have a bunch of, any need for the 30pf and 4MHz here or can we stick with the usual?
4. No issue with the pi power circuit, but any chance we could use MOSFETs we already use elsewhere on other boards?  
5. Debug LEDs. You have 1 on the PIC, but it would be nice to have the usual 3 LEDs since we have plenty of IO and 1 tends to be a bit limiting. would also be good to put 2 or 3 on the pi IO so we can tell what it's up to if we want.

Overall a really nicely layed out schematic!

zx100x100:
looks good to me, only issue is that the CAN connector needs to be rotated 180 degrees, also I'm not sure what the clearance is like, but I'm a little worried the crystal might hit the Pi

zx100x100 (changes requested):
Looks good.

A little funky that you use 0.5mm traces for 5V and GND regardless of current, vs 0.25mm for most other things but not an issue I guess.

The pin out for the AOD417 is wrong, pin 2 is not the gate. 
Otherwise looks good!

LanaTomlin (commented):
1. Is there a reason you don't have a ground plane? 
2. What the hell is the footprint for the programming header? (im pretty sure it's usually PinHeader_1x05_P2.54mm_Vertical)
3. this isn't the end of the world, but you use inconsistant trace widths. I'd understand if you used thicccer boys for ground & power planes, but it looks like it's just this little section
image
5. why dont you move the ICSPCLK via earlier so you don't have to lay a trace underneath a resistor?
image
6. under this section, there arent' any vias to the +5V plane. why don't you run +3V3 underneath that section, and then: 
* less power loss
* you dont have to run a trace underneaht Q4
image
7. barage of me being nitpicky about funcky lookin' traces:

image

image
image
image

^something about the routing direcly between two pads like that kinda ircs me but like, its fine?
instead of funky traces, can you not just,,, put more vias (circles where vias could go)
image

zx100x100 (changes requested):
Two tiny nitpicks and a comment / question. I won't stop us from ordering over these, but I'd love it if they could be fixed.
Only thing blocking is that drill files are missing.
1. every other one of our boards use platted mounting holes, I believe because they are more durable, would be nice if this board did too. 
2. Can you label the color of each LED please, typically we do DR1 DB1 DW1?
3. Just curious, why did you name your gerber folder meta? we usually call it gerbers

zx100x100 (approved):
hardware has been ordered. so one way or another this hardware is what should be on master.

we should merge ASAP, and any more problems / feedback should be tracked in an issue

dwkletke (approved):
As Zach said, board/hardware has been ordered. Approving this so it can be merged to master.

QuantumManiac (approved):
Hardware Ordered; Approving to Merge

PR #40, opened 2021-09-28
Live Telemetry v2

dwkletke:
Currently just requesting schematic reviews. Will need to cleanup/replace a few part #'s due to silicon shortage (still selecting new voltage regulators). Should be updated within the next day.

Main new changes:
-Changed radios to 1W (30 dB) ones using UART for serial communications.
-Updated LTR to new PIC with more pins, supporting USB and CAN (for future implementations). I don't have any CAN circuitry for debugging the bus at the moment, so let me know if anyone thinks it's worthwhile right now.
-Added new battery connector to LTR to support higher current requirements from radio. I am not very well versed in multiple power source design schemes, so if anyone has any input on what to add/lookout for, please let me know.

---
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/40)


zx100x100 (commented):
Looks really nice and clean, a couple notes:

1. Transmitter board looks great! only thing is I think this is old from radio board, but as nice as this paragraph it, you should probably at least list a recommended set of resistor values.
image

2. on receiver board r_sense value is missing, would also be good to select a specific part number for the Schottky diodes. 
image

Looks great! :fantasticstuff:

BluCodeGH (commented):
Also looks great to me!
1. Add values to the fuses on both boards if possible.
2. Maybe mirror Q3 vertically on the transmitter to make it a bit cleaner?
3. I liked your AP1509 layout on the receiver, maybe do something similar on the transmitter so it's easier to tell apart from the mosfet circuitry.

ZTeper:
image

No daisy-chaining please

ZTeper (approved):

QuantumManiac (commented):
Once you've addressed the above comments, you should be good from my end! :lgtm:

dwkletke:
Addressed PR comments and added all of the part numbers. However, I checked the availability of the LTR PIC and it's out of stock... I will be looking for a replacement tomorrow. Will also be starting layouts tomorrow to hopefully be on routing by Tuesday/Wednesday.

dwkletke:
Updated LTR PIC to [PIC24FJ64GU205](https://www.digikey.ca/en/products/detail/microchip-technology/PIC24FJ64GU205T-I-PT/13636094). Going to get to layout and will re-request reviews.

dwkletke:
LTT routing complete (need to add silkscreen labels still). Still need to work on LTR so just looking for LTT reviews.

dwkletke:
Also note: forgot to add a copper pour around the oscillator.

BluCodeGH (commented):
I had some time to kill so very nitpicky review, feel free to ignore stuff you don't want to fix.
1. Sorry for not catching this during schematic reviews but no jumpers to have 12V always connected like there are for 5V and GND?
2. You have a bunch of traces that overlap a pour of the same net (eg 5V traces on the top / GND traces on the bottom), the layout will be simpler to understand if you remove them.
3. You can switch around the traces to R2 and R3 to make their routing a bit cleaner (have the R3 trace on top)
4. Daisy chain on D3
5. It might be better to use a via to get 5V to the back plane rather than the pin on C9, same with GND on J2 and C5
6. You might be able to make the U3/C11 routing cleaner by moving the via to between them

dwkletke:
> I had some time to kill so very nitpicky review, feel free to ignore stuff you don't want to fix.
> 
> 1. Sorry for not catching this during schematic reviews but no jumpers to have 12V always connected like there are for 5V and GND?
> 2. You have a bunch of traces that overlap a pour of the same net (eg 5V traces on the top / GND traces on the bottom), the layout will be simpler to understand if you remove them.
> 3. You can switch around the traces to R2 and R3 to make their routing a bit cleaner (have the R3 trace on top)
> 4. Daisy chain on D3
> 5. It might be better to use a via to get 5V to the back plane rather than the pin on C9, same with GND on J2 and C5
> 6. You might be able to make the U3/C11 routing cleaner by moving the via to between them

1. I think since the 12 V line on the CAN bus is only used for a few boards, then shorting the 12 V net directly to the 12 V CAN line wasn't deemed super necessary. However I added it in just in case (good catch)! :)
2. Some cases it's to connect a pin to a pour (see pin 4 on J5). However, I think some were left from when I was iterating on the pour. Ty for letting me know (and please let me know if I missed any)!
3. I spent probably too long on the LED routing... and I still didn't do it write LOL. Fixed now!
4. Fixed :)
5. I missed this, ty!
6. Fixed :)

dwkletke:
Both LTT and LTR are in need of routing reviews now :)

BluCodeGH (approved):
LTT: You can easily check for traces that match the plane by using kicad's "highlight net" view. There is one GND trace left on the back, otherwise LGTM!
LTR: There is some weirdness with the pour and the oscillator capacitors, and with U5, Otherwise LGTM.

QuantumManiac (changes requested):
LTT Review: 
What's the point of this trace going all the way to the ground via beside the caps? Would routing it to one of the vias under the clock be inadvisable?
image

Routing could potentially be simplified here?
image

LTR Review:
I see what Jack means regarding the odd pour - are these vias supposed to be GND? 
image
Additionally, quite a few other +5V pads seem to be disconnected according to the DRC

dwkletke:
> LTT: You can easily check for traces that match the plane by using kicad's "highlight net" view. There is one GND trace left on the back, otherwise LGTM! LTR: There is some weirdness with the pour and the oscillator capacitors, and with U5, Otherwise LGTM.

- LTT: Yeah I fished it out. Thanks for catching that.
- LTR: Fixed some weird pour generation around the oscillator caps, and fixed the 5V vias to be GND instead. Also fixed pour generation on U5.

dwkletke:
> LTT Review: What's the point of this trace going all the way to the ground via beside the caps? Would routing it to one of the vias under the clock be inadvisable? image
> 
> Routing could potentially be simplified here? image
> 
> LTR Review: I see what Jack means regarding the odd pour - are these vias supposed to be GND? image Additionally, quite a few other +5V pads seem to be disconnected according to the DRC

LTT:
- A while ago on another review there was some discussion about isolating the oscillator and keeping out other traces from it's small contained circuit (can't find it atm - just trust pls). Generally speaking, I believe there are some unintended side effects that happen when routing around oscillators, as they are very EMI sensitive. I'd rather always be on the safe side and try not to route anything through the oscillator or caps themselves.
- I agree with the routing simplification. I was wondering why I routed it that way originally, and realized it was to have the U5 +5V pins on the same pour. A quick movement of traces fixed that up as well.

LTR:
- As said above on Jack's comment, these were fixed and replaced to be GND.
- Apparently DRC lied to me before I committed originally (or I just forgot to check a second time). This has been fixed as well :)

BluCodeGH (changes requested):
Jared is mad about GND traces when there is a GND pour. Also 5V traces where there is a 5V pour.
DRC is mad about DRC things.
Crossing trace? Seems to be missing some vias?
image
Zachary says you should double check the PIC footprint/pinout and the radio connector.
Alright board games are starting that's all we have.

QuantumManiac (approved):
> Jared is mad about GND traces when there is a GND pour. Also 5V traces where there is a 5V pour.

Agreed. Could probably shorten up or get rid of some traces in favor of more direct connections by via to the pours

> DRC is mad about DRC things.

Very odd, I'm not seeing anything of the sorts. Doesn't look like I forgot to pull or anything.

> Crossing trace? Seems to be missing some vias?

Something's funky, because that horizontal trace is on the bottom layer on my end.

Do double-check things as the above comment suggested. Otherwise, LGTM

dwkletke:
> Jared is mad about GND traces when there is a GND pour. Also 5V traces where there is a 5V pour. DRC is mad about DRC things. Crossing trace? Seems to be missing some vias? image Zachary says you should double check the PIC footprint/pinout and the radio connector. Alright board games are starting that's all we have.

- I thought I removed them all but I missed a couple. TY! LMK if there any specific ones that I missed again pls.
- Hmm... I don't see any DRC issues... Mind sending a screenshot?
- I don't see this. That D2-Pad1 trace is on the back layer:
image
- I have cross-checked the pinouts/footprints but I feel like it would be beneficial to get a _different_ set of eyes on that if anything. Note: there should be plenty of spacing for the radio module to slot into the connector.

dwkletke:
> > Jared is mad about GND traces when there is a GND pour. Also 5V traces where there is a 5V pour.
> 
> Agreed. Could probably shorten up or get rid of some traces in favor of more direct connections by via to the pours
> 
> > DRC is mad about DRC things.
> 
> Very odd, I'm not seeing anything of the sorts. Doesn't look like I forgot to pull or anything.
> 
> > Crossing trace? Seems to be missing some vias?
> 
> Something's funky, because that horizontal trace is on the bottom layer on my end.
> 
> Do double-check things as the above comment suggested. Otherwise, LGTM

LMAO glad we are on the same page :)

BluCodeGH (approved):
Yeah something was definitely wrong with my kicad last time.
I'm still seeing a lot of GND/5V traces on LTR
image
image
And a handful of 5V on LTT
image
Otherwise looks good, I checked the radio module pinout and I think it's right. Approving for when those traces become vias.

dwkletke:
Updated LTB with new radio connector/components. Went all the way to routing since the changes weren't super large. But feel free to review schematic and layout/routing for LTB :)

BluCodeGH (changes requested):
LTB review:
1. In the schematic, it's probably best to call it +5V instead of VCC since that's what the radio needs.
2. Any reason to have components on both sides of this pcb? Seems like you've got plenty of space.
3. You have GND traces on the back (ground) fill around U1 / C4.
4. C1 and C5 would probably be best placed elsewhere, they seem to be getting in the way of the dense routing around there. If they need to stay there, I'd suggest routing their GND to pin 1 of the radio connector rather than the via way down below.

BluCodeGH (approved):
J1 silkscreen overlaps C1's pad. Otherwise LGTM.

ZTeper:
Please fix DRC complaints regarding silkscreen going over the edge cuts.
Also, please check Digikey for various voltage regulators out of stock

ZTeper (changes requested):
https://github.com/waterloo-rocketry/canhw/pull/40#issuecomment-1023657702

dwkletke:
> Please fix DRC complaints regarding silkscreen going over the edge cuts. Also, please check Digikey for various voltage regulators out of stock

The DRC complaint for a silkscreen going over the edge is only for component J4 (radio header). This is on purpose for spacing concerns and doesn't negatively affect anything. The voltage regulator for this board is the same as the previous 2 LT redesigns, which we already ordered extra stock for in the fall. I will be using that and not looking for a new component.

ZTeper (approved):

QuantumManiac (approved):

LanaTomlin:
k this is just curiousity, but why is there a jumper here? 
image
SD1206S040S2R0 diode is out of stock on Digikey
image
those are some prettty wiggly traces
oh its on the fuse so its probably fine 
image
buck converters usually have pretttty specific guidelines on how it all needs to be routed, so it might be worth double checking if this is gonna be okay 
image
if there's a way to avoid this wiggglinggg that...would probably be good 
image
image
nitpick, but since im probably gonna be requesting changes anyways: all your power traces are slighly thicker, except for these ones


OKAY TRANSMITTER: 
image
you could probably add a pour of +12V here, and then you wouldnt need... so m many round about traces, but def not required
this buck converter looks better to me, the inductor is closer to the IC. stilll mighttt wanna look into what the datasheet says 
other than that im happy

the breakout is beautiful 10/10

LanaTomlin (dismissed):
im sorry for breaking the approve streak but im worried about your buck converters and your diodes

rdakhil (approved):
To add to Lana's comment on the buck regulator, it doesn't seem like the datasheet has any specifications on layout. but generally for a buck regulator you should minimize the distance between L1, D4, and the output caps (make that loop as small as possible) close to the output pin. Also might be a good idea to add some GND vias near/under the buck to help dissipate heat

Otherwise LGTM, gonna approve because Lana already requested these changes

dwkletke:
@LanaTomlin @rdakhil Thanks for bringing this up. I appreciate this getting sorted out now instead of us not knowing why the regulators weren't working :)

dwkletke:
> k this is just curiousity, but why is there a jumper here? image SD1206S040S2R0 diode is out of stock on Digikey image those are some prettty wiggly traces oh its on the fuse so its probably fine image buck converters usually have pretttty specific guidelines on how it all needs to be routed, so it might be worth double checking if this is gonna be okay image if there's a way to avoid this wiggglinggg that...would probably be good image image nitpick, but since im probably gonna be requesting changes anyways: all your power traces are slighly thicker, except for these ones
> 
> OKAY TRANSMITTER: image you could probably add a pour of +12V here, and then you wouldnt need... so m many round about traces, but def not required this buck converter looks better to me, the inductor is closer to the IC. stilll mighttt wanna look into what the datasheet says other than that im happy
> 
> the breakout is beautiful 10/10

So I updated the LTT and LTR based on your & @rdakhil's comments. Just some quick things to note:
- The SD1206S040S2R0 was already ordered for LTT and LTR (and is not on the LTB) so there should be no issues there.
- The jumper for the MCLR pin on the LTR is taken from this note in the datasheet (however I did replace the type of jumper):
image
image
I felt that it wasn't very clear on when to use the jumper or to just directly connect, so I opted for the jumper to be safe (since we can always have it permanently connected with that).

Other than that, I think I addressed all of your other concerns (moved the buck converter closer to its required components, fixed some of the weird trace paths and sizes, etc.). LMK if there is anything else :)

BluCodeGH (changes requested):
Your silkscreen dates could do with an update :stuck_out_tongue: 

LTT:
- You should rotate L1 180 degrees. and move the +12V trace that's currently above it to below it so you can route directly from D4 to L1.
- Why is this a through-hole programming connector?

Looks good otherwise!

dwkletke:
> Your silkscreen dates could do with an update 😛
> 
> LTT:
> 
>     * You should rotate L1 180 degrees. and move the +12V trace that's currently above it to below it so you can route directly from D4 to L1.
> 
>     * Why is this a through-hole programming connector?
> 
> 
> Looks good otherwise!

- Ty for mentioning the silkscreens. They have been updated :)
- L1 has been rotated and the area rerouted :)
- I have always opted for THT programming headers since I though they offered more stability for the constant plug-unplug... Is there a reason you think I _shouldn't_ have them besides other boards using SMD?

Ty for the review <3

BluCodeGH (approved):
Yeah I like the SMD programming connector purely for consistency. As long as we have THT ones in stock (iirc we do) LGTM!

ZTeper:
image

The traces are a little close to the drill holes. Make sure this is within the manufacturer's tolerance.

Artem-Sotnikov:
Breakout board comments:

Looks good overall, avery well-designed board. However, I have a few comments on clearances with holes. We should probably avoid putting the traces in too close to the vias and holes, especially since in these cases they can just be moved further aside with no issues. 

See the trace next to the RX_H through hole:
image

See the GPIO traces next to the screw hole
image

Artem-Sotnikov:
Comments on Receiver board:

No comments on schematic, very well designed and everything checks out. On the PCB, everything also looks good. However, I agree with Zachary's earlier note on clearances. Furthermore, considering that the board will be hand-soldered one other particular spot is of possible concern. 

The through hole on the battery fuse is a bit too close to the ground pad of C3. Since this is the live end of the fuse going to the battery, a short between these two points could be very bad. We're not really strapped for space here so I think it could be moved aside with no issues.

image

Artem-Sotnikov:
Comments on transmitter board:

Schematic and PCB look great. A few minor points. 

Odd silkcreen labels for 3v3 and !MCLR on J1, is there any reason why they're not placed directly next to their pads?
image

Another clearances nitpick on the 3v3 line in J1
image

Silkscreen says U6, but it is actually u3?
image

Artem-Sotnikov:
Overall, everything looks really good and all of my comments are very nitpicky (save for maybe the silkscreen mismatch on the transmitter). I'd say that addressing the rest is up to the author's discretion.

Artem-Sotnikov (approved):

dwkletke:
Thanks for the reviews!

- I think I fixed all of the clearance issues mentioned (and a few additional ones I noticed).
- For the silkscreen label on the programming header on the LTT, I think that was from previous components blocking that space but I moved the lables back now.
- I think U6 on LTT is properly labelled (at least when I checked). What makes you think it isn't?

Artem-Sotnikov:
I think I mixed up the schematics with U6 (that's what I get for having a billion kicad windows open). Everything looks great now; no more comments! Let's order this thing!

LanaTomlin (commented):
nitpick but you can prob. reduce the size of these spots where ground pour is missing:
image
image
you might wanna bring c5 closer, also might wanna use a thicker trace size for your power lines .

Also the same as last time, just wanna confirm with y'all that this layout is close/similar to the one reccomended in teh datasheet?

also i fucked up and woke up late will finish this review as soon as I can I'm really sorry

BluCodeGH:
Ordered, merging.

PR #42, opened 2021-12-06
USB Debug Respin

BluCodeGH:
Adds silkscreen labels to CAN connector and an optional terminating resistor. Yes I know the trace hops are ugly but I don't want to reroute more than I already have.

---
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/42)


QuantumManiac (approved):
LGTM. Update schematic PDF once everything is good to go pls.

Quick google search says that the resistor should be 120 Ohms and because I don't know any better, I'll take your word for it.

Before and after pics because it was helpful for me:
PCB:
- Front
image
image
- Back
image
image




Schematic:
image
image

zx100x100 (changes requested):
Looks good, only thing is it would be nice to have labels on the jumpers so you can tell which is which

rdakhil (approved):
Looks good to me!

Artem-Sotnikov (approved):
Overall, lgtm. 

A few notes:
- Making the text point all in the same direction could be a good idea, especially since the board is not particularly tight on space
- Maybe on some of the more sensitive signals, stitching the ground planes might be a general good practice sort of thing. Although, this may be my emf noise paranoia speaking.

zx100x100 (approved):
:lg::tm:

Issue #30, opened 2020-10-15
Mini sensor issues


zx100x100:
1. The PTC fuse values should be labelled

2. this board could really do with some SMD test points.

zx100x100:
MCLR silkscreen does not indicate MCLR voltage

rdakhil:
some ground test points would be nice to have in the next respin

also standoffs are there in the CAD but they are very thin. the plating should be changed to be about as thick as the mounting holes

JasonBrave:
Clock source(oscillator) is not isolated, result interference from CAN, cause PIC think external oscillator failure.

BluCodeGH:
Recent comments are not addressed.

Joe-Joe-Joe-Joe:
Has this board not been respun since this issue was raised?

PR #44, opened 2021-12-10
IMU changed on MiniSensor

eliaskountouris:
Changed IMU on minisensor board to MPU-6050. 

Datasheets:
MPU-6050: https://invensense.tdk.com/wp-content/uploads/2015/02/MPU-6000-Datasheet1.pdf

---
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/44)


BluCodeGH (changes requested):
Weird chip but this looks right comparing to the examples.
1. Having pin 20 (NC) connected to ground is weird, probably best to rename it RSV (reserved) like they have or leave it unconnected.
2. Better to stick with a local-only netlabel (button above the global one) rather than a global netlabel for nCS.
3. Good to add little text notes saying what the new ICs are, just like the barometer and other IMU have. It's not clear atm that part of this is 1V8.
4. It would make me happy if R1 and C14 were lined up with the other passives there, generally feel free to spread things out so they look nice. The goal with a schematic is to communicate, so good to prioritize prettiness over compactness (not that this is overly squished, just something to keep in mind).

QuantumManiac (changes requested):
So far so good, just a few minor nit-picks for aesthetics purposes:

It would be nice if the labels for `C16` were to the right of the components, like the others beside it:
image

Perhaps add a bit of wire between the intersection and the ground over here? Feels a bit cramped:
image

I feel like this entire bit could be a bit less cramped. Like Jack mentioned, compactness doesn't matter much but communication does:
image
This bit could be moved a bit more to the right so the `GND` and `3V3`  labels aren't so close. I also just noticed that the `3V3` label isn't centered:
image

QuantumManiac (commented):
LGTM, should be good to proceed to layout once there's another review or two down

BluCodeGH (approved):
LGTM! Tiny nit, I'd move the +3V3 connection up to be level with the other wire, but I think you're probably good to proceed with layout.
image

QuantumManiac (approved):
Marking as approved so Toast stops bugging me. Not actually approved - there's PCB layout to be done.

BluCodeGH:
While you're working on this board you should try to resolve #30. In terms of where we'd want test points, @zx100x100 do you remember what you were wanting?

BluCodeGH (changes requested):
1. Wow DRC really hates this board, doesn't it. Looks like some of the complaints (courtyards overlapping, traces too close) you should be able to fix. If it's complaining about random parts of traces maybe try redoing them? I think some of those traces are needlessly close, you should be able to spread them out.
2. DRC also says you have a few unconnected items. Remember that even if things are connected to the +3V3 or GND pour, those pours all still need to connect. For example there is a bit of isolated GND pour here, and similar problems elsewhere that DRC points out.
image
3. If there is a pour on a copper layer, there should be no traces of the same net on that same layer. So in this case, there should be no GND traces on the back or +3V3 traces on the front. These issues are easy to spot with the 'highlight nets' tool.
4. The back layer is not nearly as stuffed as the top, I'd suggest putting some more components there to give yourself some breathing room.
5. Try to avoid traces intersecting at right angles.
image
6. There is a whole lot of daisy chaining going on with this board. Probably good to spend a bit of time here and fix it.
7. I think it's best to prefer vias to traces when it comes to routing something that's poured on the other side. For pads super close to each other sharing a via is fine but even something like this is a bit far imo.
image
8. C11 is too close to U5. We need room to get tweezers and a soldering iron in there at minimum, and the more space to work with the easier it is to solder.
9. Watch out for extraneous bits of wire like on the +3V3 pad of C11, they don't end up making a difference (the entire pad is copper) but clutter up the schematic.
11. You have various overlapping silkscreens, eg C8 and C14. Try to keep silkscreens out of courtyards and definitely off of pads. Also keep them off of vias if possible.
12. Feel free to choose a different test point footprint if you need the space, but definitely try to fill up the back of the board first.

zx100x100 (commented):
Seems like jack had lots of feedback so I'll let you address that before doing my own review of the layout, but schematic looks good. only thing I noticed is that there are a bunch of unwanted project files kicking around and it would be good to delete these unwanted ones. 
image

BluCodeGH (changes requested):
Wow this is one packed board, great work making it fit.
1. Zones need a re-fill, you can do this quickly by pressing 'b'.
2. I'm still seeing +3V3 traces on the front and GND traces on the back. Make sure you're re-filling the zones frequently to fix the rats nest lines for things that should be connected with a fill. Also you still have plenty of long +3V3 and GND traces on the back / front respectively that should be broken up into separate short traces with vias.
3. Test points 1-4 should be flipped so the silkscreen is on the front face.
4. The J4 silkscreen label looks like it corresponds to U3, maybe move it so it's clear what it's talking about.

rdakhil (commented):
Good work! Just a few comments, mostly for aesthetics:

1. This area is a bit crammed together, which may cause you some issues while soldering
image

2. Doesn't really matter since there's a pour anyways, but you could clean up the that ground trace below U1 like this: 
image
 
3. This long ground trace could be removed by placing GND vias directly next to C18 and U2. Also, the GND pin on the J1 connector doesn't need a trace/via since it's already connected to the GND pour.
image
You can place as many vias as needed, so use them when possible to avoid long traces.

4. This also doesn't affect the function of the board but I would recommend removing these ground traces. You don't need any traces connecting the GND vias together because they are already connected to the same GND pour.
image
There's a few more instances of this on the board, so I would recommend going over and removing the extra traces. This might free up space for routing other traces. As Jack mentioned, you can use the 'highlight nets' tool to spot these.

5. This trace could be cleaned up:
image

6. Nitpick but this trace is sticking out a bit here: 
image

7. C10, U7, F2, J1, and J2 have silkscreens on top of vias. Try to fix them if possible.

BluCodeGH (changes requested):
Getting pretty close!
1. Still missing some front 3V3 traces and this massive ground pour
image
2. J4 is the 3 pin connector in the corner, that silkscreen label position is incorrect.
3. Stuff here is still really close together. Take a look at the spacing between C10 and C8, I wouldn't put components any closer together than that in the "short" direction (remember we need to fit tweezers in there). In the long direction R9 to C10 is pretty tight but we can make it work. A lot of your components on the back are way too close in the "short" direction.
4. Why is there an arrow on the U6 silkscreen?

LanaTomlin:
1. schematic is missing NC symbols, if a pin isn't connected you should end it with a littel blue x
image
2. looks like you didn't run ERC, might wanna do that
3. 
image
these connections look....really messy if you could clean 'em up, that would be nice
4. it might be an idea to add a bit of text to explain the thermisotr and tank pressure connection, cause it's not explicitely clear how that works
5. what's a "200mR" resistor, and did you do the calcs to make sure you want 200mΩ, if so can you put 'em on the schematic
6. linear regulator wirign looks kinda messy, would be nice to clean it up
7. UHHH i cant open the pcb idk if thats a me problem or a you problem

image

also if you could...maybe cleean up the bajillion filres in this folder that would be nice:))
k um thanks for your patience in the literal two months it took me to do this review, and uh hopefully ill review the pcb soon gottta RUNNNN

LanaTomlin (changes requested):
UGHAS;D L im bad at git

BluCodeGH (commented):
2. All of the stuff ERC is complaining about is just about the power flags and the NC flags being missing (NC is an easy fix)
3. I wouldn't worry about cleaning the schematic up too much at this point, we want to get this board in within the next few days.
4. Likewise
5. 200mR is 0.2 ohms, thats what the current version of mini sensor uses and is fine.
6. See 3.
7. @LanaTomlin you need to update to kicad 6.

LanaTomlin:
I would argue that it takes like, 10 min, to clean up the schematic and its a pain to get a pull request through, so its a lot easier to fix the schematic *now* rather than wait until later and have to do this all over again

BluCodeGH (approved):
Yeah that's fair. Apart from Lana's comments PCB LGTM!

A few cleanup tasks:

- Please remove the old kicad 5 files (.pro and .sch, along with the .sch-bak).
- Remove the old .pdfs, generate a new one.
- Add the kicad backups folder to the .gitignore (add the line `*-backups/`).

ZTeper:
DRC is complaining a lot; but most of these complaints are not critical issues. Please fix:

- hole clearance problems
- drill hole size issues
- silkscreen going outside the board edge cuts

Please also give meaningful silkscreen labels to the testpoints, not just "TP1", "TP2", etc.
Why is J3 on the reverse side of the board? Would it be difficult to plug in a cable to it?
Please mark the MCLR of the 5-pin programming header clearly with a silkscreen label

ZTeper:
Please add a large decoupling capacitor between +12 V and GND, otherwise your +12 V rail will be rather unstable

ZTeper:
Please also label the function of each pin on J3, J4

ZTeper (changes requested):
#44

BluCodeGH:
> * drill hole size issues

Those tiny holes are part of a predefined footprint and within JLC's capabilities, I guess change the DRC rules if you want this to go away.

> * silkscreen going outside the board edge cuts

I'd be inclined to leave this, we already know everything fits since this is a revision.

> Why is J3 on the reverse side of the board? Would it be difficult to plug in a cable to it?

Looks to me like otherwise it will conflict with J4.

eliaskountouris:
@LanaTomlin @BluCodeGH Everything is good now, with the exception of 4. The therm / pres connection stuff was on the original version, and I'm not completely confident I know how it works. Could you provide me a run down so I can add some text on to the schematic. Thanks.

After that I should be good to push.

BluCodeGH:
The thermistor forms a voltage divider with R9, we measure the resistance via that voltage.
For tank pressure we get a voltage on pin 2 that we scale down on the resistor divider.

eliaskountouris:
Also can someone double check if C17 is a large decoupling capacitor / tell me how to properly calculate the size. I just used the same capacitance as C19 (+15V decoupler)

BluCodeGH:
Given that +12 will be a battery rather than a switching supply that looks fine to me. 200mR is what the team usually calls 0.2 ohm resistors, I'd stick with that for consistency (and the ohm symbol has a chance of not displaying correctly, you can bet 200mR shows up right).

ZTeper:
The routing to the crystal could be improved. 
image

BluCodeGH:
The lower suggestion there would constitute a daisy chain on the capacitor. The upper one would be fine since daisy chaining isn't an issue with through hole components.

LanaTomlin:
image
are we sure we're gonna be able to squeeze boththoseconnectors on there? i usually like to leave space between them for, like tolerance 
also apologies for not full review (still) i need to zoom to work

BluCodeGH:
Yes they'll fit (one is on the back), provided they haven't moved since the last revision of this board.

LanaTomlin:
Okay so I haven't looked at it supppperr thouroughly, but it looks good to me. 
The only thing I'd say is that CANH and CANL are suppposedd to be differentially routed (we usually don't bother), but these do def have different trace lengths. I don't see a super easy way to fix this, and I don't think it's critical so I'm gonna approve anways. (If I can figure out how)
image

LanaTomlin (approved):

rdakhil (approved):
LGTM

QuantumManiac (approved):
Could probably move this label down to make things look a bit less squished:
image

Otherwise, LGTM

ZTeper (approved):

BluCodeGH:
Ordered, merging.

PR #49, opened 2022-05-28
Fix voltage divider resistor values on LTT

BluCodeGH:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/49)


dwkletke:
@BluCodeGH  Can you clarify what is being changed here? I don't see any difference between resistor values... But I might be blind 😅

dwkletke:
> @BluCodeGH Can you clarify what is being changed here? I don't see any difference between resistor values... But I might be blind 😅

Was using the wrong kicad version... My bad 😄

dwkletke (approved):
Technically the comment stating the range of the divider's output voltage is _slightly_ outdated (0-2.031V) but you can update that if you'd like.

Rio6:
closing this as ltt got a new revision and I dont think this applies anymore

PR #51, opened 2022-11-15
Charging Board

jasonskells:
Functionally this schematic can be reviewed, there are only a few small things listed below that need to be decided on.

Removed the LTC4414 Ideal Diode Source Selector IC (~10CAD) and replaced it with diode OR circuit. If power dissipation is a concern I can swap the diodes with larger ones that can dissipate more power, or worst case put the ideal diodes back.

Questions: 
Is there anything else that we want to be able to report to the can bus? Currently the battery charge status, battery voltage and the 13V rail are being reported. Do we want the 5V rail to be measured as well or is that unnecessary?

Do we want any fuse protections on this board?

What connectors do we want to use for everything?
 - Screw terminals or does @Artem-Sotnikov have recommended connectors that I should use?

Do we want to be able to control when the battery is charging?
 - This just involves sending a GPIO to the CE (charge enable) pin on the BQ24600 battery charger.

Do we want any indication LEDs?

Do we want to use the over temperature protection feature of the BQ24600? 
- I think it's intended for a thermistor located on a battery pack, but we could place it on the charger board and have some temp protections. Downside is this will always be consuming a small amount of power unless we place a FET in its path.

What current do we want to charge the battery? Currently, the BQ24600 is set to charge at 1A.

I need to revisit the BQ24600 input filter and do the calculations to confirm suitable values. The 2R and 2.2uF cap are just taken from the example application schematic in the datasheet.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/51)


BluCodeGH (changes requested):
Looking great overall, still some work to do tho.

I'll first note that architecturally there are some tweaks to make. The biggest issue here is that you are missing our standard harwin connector. This is the super robust 6 pin connector that all our boards have, which connects them to the CAN bus. Most boards use 5 of the 6 pins: GND, CANH, CANL, +5V and +12V (unregulated, aka +13V). In this case we can use that sixth pin as the 24V input from the ground, and this is the connector where we should output a stable 12V (unregulated, the VCC line in your schematic) and switched 5V.

The only other connectors on this board should be a ring terminal block for the battery (see live telemetry transmitted for the part no / footprint) and a connector for the radio GPIO. Screw terminals aren't an acceptable connector type for flight boards, we could either use another harwin connector (expensive) for this or maybe just a JST or similar. Artem probably has more ideas.

Responding to your notes/questions in order:

With the diode OR, won't we be drawing power from the battery even when plugged in? I don't think the charging will work properly if we're also drawing power at the same time. It's pretty simple to make a preferential power selector with a mosfet, but if we can use a powerpath switcher that guarantees the output won't see a momentary blip when switching that would be ideal.

We want to be able to report battery charge status, battery voltage, and current draws on both the 5V and 13V lines (ie reporting total current draw including ourselves, but also the rest of the CAN bus). Reporting the 13V and 5V voltages can't hurt either.

Yes, all our boards should be fused. Typically we use PTC resettable fuses, but given the higher current draws associated with this board I'd recommend a proper fuse off the battery connector. See the current version of live telemetry transmitter for the correct fuse holder footprint to use.

See architecture comments re connectors. Copy the harwin connector from any canhw board.

I don't see a need to control the battery's charging. Can't hurt to break out a GPIO with the proper pull up/down tho.

All boards should have 2-3 debugging / status LEDs hooked up to the PIC.

I don't think we need to worry about overtemperature provided we choose a sufficiently low charge current. We aren't really in any rush to charge at all, its more about keeping the battery topped up and not draining it throughout the day.

On that note, I'd probably reduce the charge current to 100 or 500 mA (those are the two options on adafruit's lipo charger). If its just a matter of swapping the divider maybe design for 100mA to be safe and then we can up it if deemed necessary in testing.

A few initial comments on the schematic, although I'll do a more thorough review when the architecture stuff is worked out:
1. Please choose a more descriptive rail name than +VSW. We can make custom power symbols, maybe do that for +13V.
2. You can use a second mosfet in parallel with Q3 for independent switching, then either mosfet can pull down Q4's gate.
3. It's probably easier to use a fixed 12->5 regulator than a variable buck, either is fine tho.
4. Please actually set the values for your R_sense resistors, if you don't then its up to the person assembling the board to figure out an appropriate value and then the value isn't documented for the person writing the code. You should size it such that the 100V/V INA180s give an acceptable output voltage range (PIC's max analog input should be 4.096V) at the expected currents.
5. You're missing your CAN transceiver.

BluCodeGH (changes requested):
Looking good! Some notes:

1. J1 will be wired directly to the electrical disconnect so it doesn't need +13V or +5V_SW. Also good to note this difference between the connectors in a text note on the schematic.
2. A lot of your net labels have their "connection point" on the wrong side so the wire has the little disconnected square. Here I've fixed the CANH connection by flipping the net label (x), see the difference between the non-flipped CANL
image
3. The CANH/CANL wiring of the transciever looks like it got messed up.
4. Looks like you made this with a 0.025in grid, that's fine but generally try to stick to 0.05in please.
5. Per the long thread in #electrical we're removing the GPIO to the radio.
6. That ideal diode MOSFET package will be hellish to solder, are there better options?
7. LTC4414 datasheet says you can add decoupling capacitors to VIN and sense, probably a good idea.
8. 0.2 ohm current sense resistor * 2 A * 100 V/V amp = 40V. I'd recommend dropping it down to maybe 0.02 or less so we can measure 2-3A.
9. Any chance there is a battery charger IC with a better package? We'll probably need to reflow this one which is a pain.
10. Looks like you have SRN and SRP wired backwards.
11. I think your charging current math is wrong? I get 50 mA?

jasonskells:
1. Fixed.
2. Fixed, didnt realize they had to be connected to the end of the wire, so thank you!
3. Fixed.
4. Thought I had stuck with the 50mil grid, I'll make sure to fix this going forward. I changed the grid to 50mil and selected everything and clicked align to grid so I think it is fixed now though (although some of my symbols might not be, I'll fix these too).
5. Fixed.
6. I don't think this will be an issue to solder, I plan to reflow this board anyways.
7. Fixed. Added another cap on the output that is a DNP so we have the option to populate in case we need more capacitance.
8. Fixed, changed to a 0.02R 1W resistor.
9. I think soldering the BQ24600 will be fine, I don't really have an issue with QFNs, just need a hot air reflow station/hot plate/ oven, and a stencil for this board (I'll pay for the stencil if the team can't/won't 😅). The **_only_** other option I've found which is easier to solder is the: MAX1873SEEE+. Although this isn't explicitly for a LiPo, a Lithium Ion battery has the same charging method so it should work. I would like to stick with the BQ24600 but if you think it'll really be an issue, I can change it/look deep into alternatives.
10. Fixed. Yes I did, thank you for pointing that out.
11. Yes, I only wrote R2/R1 instead of R2/(R1+R2). This has been changed since I've changed the current sense resistor to 0.02R. The math has been fixed now. 

Thanks again for all the feedback!

BluCodeGH (approved):
Looks great, I found a bunch of nits but they should all be straightforward to fix so feel free to proceed with the PCB.

1. I was talking to Zach and he convinced me that it's better to use the unused pin 6 on the main CAN connector for +24V in, since its not very much harnessing work and will make the electromechanical integration easier. So we'd remove J1 and just have J2 but with pin 6 connected to +24V.
2. We use 0805 for everything (and actually 1206 for the LEDs). It's better to be consistent here so parts can be exchanged and the board can be assembled by more people.
3. If we want those LEDs to be visible in the desert we better give them 10-15 mA. Also they should all be different colors (we typically stock red, white, blue and green so pick your favorite three).
4. Probably easier (both assembly and code-wise) to make R5 and R7 the same value.
5. J2 should be connected to +VCC (always on) not +13V (only powered when we are receiving 24V) right?
6. Right now your 13V rail CSA maxes out around 2A, that's probably a bit low. I'd drop it down to a 0.01R shunt to have a good margin.
7. When I saw the `56k2` label on R18 I first thought you had a typo and meant `562k`, on the team we would typically denote it `56.2k`. Likewise for R11
8. You should specify that R12 and R13 need to be 1%.
9. CSSP and CSSN should be shorted to DCIN according to page 11 of the datasheet.
10. I think Vbatt in your C13 calculations should be 12.6V (3 cells) which slightly reduces the capacitance requirement. At a glance it might be hard to hit that ESR requirement with a single cap, you maybe want a few in parallel?
11. The power sheet hierarchical pin CURR_VSW was renamed in the power sheet but not on the main sheet (you should be running ERC and disabling the pin types warnings to catch stuff like this).
12. Looks like your custom footprint library isn't actually in the project somehow? Idk what's up here, take a look at it pls.
13. You can switch the programming connector to the surface mount version if you want.

rdakhil (approved):
Looks good to me! Just one small thing and a few nits:

- I think the PFET on the battery charger IC EXT pin (Q2) is flipped. VIN should be connected to the source of the FET instead of the drain, as shown in the circuit on pg 9 of the MAX1873 datasheet

nitpicks:
- there are a few things it would be nice to note on the schematic:
     - the voltage range of "+BATT"
     - ik it's on webench but the max output current on the power regulators would be nice
     - all inductor part numbers, so that we can quickly confirm their footprint/package, inductor saturation current, and availability
     - the CSA gain

BluCodeGH (changes requested):
Points 3, 4, 5, 6, 9 of my previous review and none of Randy's points have been addressed. If our reviews are mistaken please leave a comment explaining why.

BluCodeGH (commented):
PCB looks great, I'm really impressed by this layout., One thing I spotted is that this pour between Q2 and L1 is a bit cut up:
image

We talked about this in-person but I'll write it here for posterity, the biggest issue I see with the PCB currently is hand-solderability. I get nervous when I see two pads directly facing each other, since it means you'll have to solder one of them first and then it will get in the way of soldering the other. This is particularly an issue around both sides of L2, interfering with U4, R23 and U7 specifically. Rotating those interfering components by 90 degrees so their pads don't point at the L2 pads would solve the issue.
image
Similarly in the top corner Q2 and C7 will be tricky, as will R11 / R13 and R12 (unlabled) / C8. For the resistor set points you can definitely afford slightly longer traces so I'd consider moving Q1 down a bit, moving R11 to follow it, and shifting R13 down and R12 left. That said there are probably a bunch of configurations that work out nicely so see what you find that works.
image

rdakhil (commented):
1. Q2 is still not connected correctly, the source should be connected to VIN
2. Daisy chaining between R9/Q1
3. More daisy chaining around R10/R11, R18/R19, R20/R21, R24/R25/Q5, and some more places
4. As jack mentioned, a lot of these components are very close together and may be difficult to solder. Especially those components with solid pad connections to copper pours and no thermal reliefs

jasonskells:
1. Fixed
2. I dont think this should be an issue. 
3. Fixed most of these. 
4. Fixed, spaced out components and adjusted orientation to make for easier soldering

BluCodeGH (approved):
I'd update the schematic to make the debug LEDs different colors (your options are 3.3V blue and white or 2V green and red), PCB LGTM.

on_pad_charging/README.md
```diff
@@ -0,0 +1,10 @@
+# On the Pad Charging
```
Rio6, line 1: why is this file in its own directory lol

JekkuChen (approved):

PR #52, opened 2022-11-17
Updated LTT to only provide power to its own radio

Rio6:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/52)


Rio6:
圖片
The current sensing should be after the inductor, replacing the +5V like this so it is the output of the whole block. The current sensing resistor is supposed to have all the 5V current going through it so we can calculate the current using V = IR

圖片
Sorry for being unclear, but the radio gpio pin would be on a different connector not on CAN bus. I was meant to say it can use the same type of connector. Also, maybe name it to something like POWER_BOARD_CTRL so it's more clear on what it does.

Just had another thought, it might be good to have another radio pin to control the 3V power using a mosfet. This way we can reset the LTT's MCU as well in case we want to reset it remotely as well.

BluCodeGH (changes requested):
Looking good! Some notes:

1. I believe the current revision of the board actually had the radio connector wired up wrong so the radio had to be flipped upside down, which made it stick up above the PCB. You should be able to work out the correct pinout from the RFD900x datasheet, maybe ask someone in loo to help if its not clear.
2. You've made `RESET_CTRL` a local netlabel so they aren't actually connected, you should use a hierarchical label (like TX) instead.
3. Stray wire off to the right of the radio module connection
4. The bootload short jumper can be removed, we can't exactly flash firmware while the radio module is plugged into the PIC.
5. I'd move the CAN bus connector onto the main sheet, and change the CANH and CANL labels to netlabels.
6. You should be more specific than "AP1509" with the part number, AP1509 has a bunch of different models with different output voltages.
7. Current sensing is wired up backwards. Currently the flow of power from AP1509 to the 5V rail is this:
image
But notice the + and - on the INA180, + needs to be wired to the high voltage, aka before the resistor. Also having the current flow from right to left across that resistor is very nonstandard and so makes the schematic hard to read. I'd recommend getting rid of `CURR_SENSE` and wiring the INA180 circuit directly up to the AP1509 output, with the +5V output on the right side of the current sensing resistor.
8. Pleas specify the value of R_Sense

BluCodeGH (changes requested):
1. Still have that stray radio wire
image
2. Make sure to annotate the schematic
3. The MCLR diodes you selected aren't hand-solderable. They also have a 1V drop which means the MCLR pin will only go down to 1V, you should be able to find diodes with drops closer to 250 mV.
4. If you look at the radio connector footprint on the PCB you can compare its pin numbering to what the radio module needs - they are different. I'm pretty sure if you rotate the schematic connector 180 as shown it's correct.
image
5. Is there a reason you chose the pin you did for the reset GPIO? I'd probably use P1.0 to make it easy to remember.
6. R3 should be labeled "0.03R" not "0.03" so we're explicit about the magnitude. I'd also leave some margin in the output, maybe go for 0.02R instead.
7. You should probably power the INA180 with 3.3V so it can't accidentally damage the pic with too much voltage.
8. U3 should use the manufacturer part number, not the digikey one. 
9. The U3 NC pin should be left floating (aka not connected), not tied to ground.
10. Also on the U3 symbol you should use the actual pin function (PG), not how you intend to use the pin (NC) so the symbol can be reused. Also also also
11. Also also is U3 the same LDO that the old version of live telemetry used? We've had issues with models that are susceptible to RF noise in the past, want to make sure we use known good ones.

Rio6:
SPX3819M5-L-3-3/TR is the known good one. The old telemetry didn't use the one that had issues but it's nice if all boards use the same one anyways.

BluCodeGH (approved):
LGTM, you should be good to move onto the PCB.

In general, an input capacitor of 10u is standard. For the others, the datasheet does actually tell you the suggested values:
image
(we have no reason to use an expensive tantalum cap)

BluCodeGH (changes requested):
Schematic looks good, PCB comments:
1. Rounded corners are sexy
2. Pls use plated mounting holes
3. I don't see your names on the silkscreen, nor a cute name for the board
4. C5 should be close to U6. I'd also rotate C6 90 degrees so its GND pad is close to the bunch of GNDs on U6.
5. Likewise D4 and L1 should be close to U6 and you should use a fat trace to connect them all. 
6. C9 should be close the L1.
7. Your 5V trace width is a bit borderline for the 1A the radio might be using, I'd up it to like 20 mil.
8. 5V is daisy chained through U6
9. C11 5V should have a shorter trace to U3, its pretty roundabout rn. It could also be closer (especially it's GND pin)

The rest of the board looks good, nice work.

Rio6:
1. Sexified the corners ;)
2. Fixed, I didn't realize they are non plated
3. Most of the design were not from me, I just did the layout. Put the old name back though.
4. The space is kinda small to be able to freely place these bulk components. But I think it should be ok since it is a bulk cap and it's not THAT far away.
5. Moved U6 a bit closer to L1 and D4 though. Trace is also fattened
6. Same as 4, also I don't think the added inductance is that much compared to the 47uH anyways
7. Fattened also
8. Redid U6 layout so it's fixed
9. Idk what was up with C11 but it's fixed now

jasonskells:
Schematic Notes:
image

-  should be a schottky diode symbol instead of a zener diode symbol

- would be good to mention on the schematic what the current limit of the buck converter is.
- change the 100uF caps symbols to a polarized, electrolytic one 
- Is a 100uf cap necessary, where did you decide on that value?   

- is the programming connector connected properly, its reverse order from other boards.  


Everything else looks good to me, it was clean and easy to read so nice job on the schematic 👍.

Layout Notes:
These are pretty much all nitpicks, use your discretion if you want to make the change or not.  
image
- I would recommend changing this trace on the Q1 drain. Also, why the two traces? Can the current of that pin not be supported with just trace?

image
- I gave more details on the other board I reviewed, there are similar potential acid traps on this board, like with the relay board you respun. This trace can be rerouted to look more like: 
image
- there are other traces around the board that can also be cleaned up if you have the time

image
- is there a specific reason we have so many vias around this crystal? Is it being used for shielding?

If you have any questions, feel free to send me a message.

Rio6:
I didn't do the schematics so can't comment too much on that. The programming header does look backwards, although the connector is just a pin header so both way works, and I already routed it this way so :P

I agree routing them througb pad looks much cleaner, except everytime I do it I get yelld at for daisy chaining

The vias are suppsed to shield the crystal yes

The split power is there because fat trace can't squeeze through the pins, and I don't want to have one point that has less current capacity that all others. I think I can prettify it a bit.

BluCodeGH (approved):
LGTM

QuantumManiac (approved):
LGTM!

jasonskells (commented):
LGTM!

jasonskells (approved):
LGTM!

PR #53, opened 2023-01-09
Mini sensor respin

Artem-Sotnikov:
Behind schedule and over budget, but the schematic is fully cleaned up to be properly reviewed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/53)


BluCodeGH (changes requested):
Overall looks good, a bunch of small things:

1. F1, F2, R31, R32, C1, C2, C3, C4, C5, C6, C7, R1, R2, R3, R4, D1 need values.
2. Why are all your test points labeled "5000"? The symbol "value" should be a description of what the test point is, eg "+5V" or "12V current amp". Also can you pls use the default kicad testpoint symbol for consistency? (Fancy keystone footprint is fine).
3. Your library is called "keysight 5000 testpoints" but the company is keystone lul
4. I'd generally just put the bourns part number as L1's value rather than "6.8 uH", then we can search for an equivalent one if it goes out of stock later but at least there's a default.
5. The `SPX3819M5-L-3-3/TR` is our "known good" 5v->3v3 regulator in a RF-interference sense so let's use it.
6. For some sensor types (eg 0-10v transducers like our pneumatic one) we want a 3-pin connector: power, data and ground. For thermistor-style applications we can just hook up the first two of those.
7. Will the value of the matched resistors change based on the application?
8. I'd label R18 and R20 as the gain resistors, and pls include the formula mapping input to output voltage.
9. CANTX and CANRX should be on pins C0 and C1. Look at another CAN board for which goes on which.
10. You have spare pins, maybe add two more LEDs? We usually like to have 3.
11. You need the PIC18LF, not the PIC18F.
12. U7 has a minimum output current, maybe we need a resistor to ground?
13. I believe the U7 drop-in replacement needs 5V in? So we should add two 0R jumpers to select?
14. U4 wants a .1u decoupling cap
15. Reminder to check for availability of mating JST connectors / switch to a different series if we can't find much.
16. Barometer datasheet says CSB must be tied to VDD or GND in I2C mode.
17. Hooking up the INT pin of the imu is probably a decent idea, there's a chance it's useful and we have spare GPIO.

Artem-Sotnikov:
Wow this is really odd, idk how a bunch of the part values got scrubbed.

Artem-Sotnikov:
For F1, F2, and D1, which parts (and thus values) do we usually use? I can't find the values in the old schematics. I could source new components and determine values from first principles but I think this would not be the best use of time.

Artem-Sotnikov:
The numbers do not add up for the 3-pin connectors; one side cannot accommodate 4 of them. It's going to be 2 3-pin and 2 2-pin connectors.

BluCodeGH:
We stock 350 and 750 mA PTC fuses in 1206. LEDs are 2V red and green, 3.3V white and blue.

Artem-Sotnikov:
I found a green LED that runs off 3.3V. For the purpose of standardization, I think all the LEDs should run off 3.3V. (The lineup would thus be blue, green, white)

https://www.digikey.ca/en/products/detail/kingbright/APHCM2012ZGC-F01/1747497

BluCodeGH (changes requested):
Almost there:

1. The new 5->3V3 LDO wants a 2.2u cap on its output.
2. IMO consistency with the voltage drops of LEDs in our green LED drawer is more important than one resistor.
3. See points 7 (just curious), 8 (formula), 9 (check actuator board), 12, 15 above

JekkuChen (approved):
Other than what Jack has already mentioned, it looks good to me

Artem-Sotnikov:
I don't think the 1 uF vs 2.2 uF on the LDO matters much but I'll change it

Artem-Sotnikov:
Also I think we're fine on the JST connectors lol 
image

Artem-Sotnikov:
Regarding the remaining outstanding points:
7. Yes, the resistors could conceivably change with the application. Maybe it is acceptable to use the same ones for varying resistances on the other half of the bridge, but this would take research to answer; it’s best to keep things flexible for now. The vishay series of matched resistors has numerous values that we could use depending on the application
8. The document linked at the top of the schematic should be consulted for the hookup guide, trying to summarize it in the schematic would make it too cluttered and hard to understand. If a separate document summarizing it is desired I could do that separately.
9. Done
12. Adding a resistor "in parallel with the load" is the approach suggested in the datasheet; but as mentioned in an earlier comment that will make the board significantly more power-hungry
15. See comment above (lots of availability), and I will try to get through to hirose. Barring that, if I have some spare time I might dig through my selection notes and see if there are any connectors which have pre-crimped cables on sale. However, I'll proceed to routing without this, since either way the dimensions shouldn't need to change

P.S. (also had to rebase my changes, should have done that earlier tbh)

eliaskountouris:
LGTM

eliaskountouris (approved):
^

BluCodeGH (changes requested):
Looking good, some layout issues I spotted and some routing nits:

1. Rounded corners are sexy
2. What are the "targets" in the top left corner on the edge.cuts layer?
3. Is there no way to fit the programming connector all the way onto the board?
4. Similarly if we can get more of the JST connectors onto the board then we can stake them down for more security, right now those look pretty easy to break off.
5. Missing solder-on standoffs for the sparkfun daugher board.
6. R31 and F1 silkscreen labels are swapped
7. TP18 overlaps with U7
8. Daisy chaining through most of your decoupling caps, eg C6, C7, C2, C10, C1, C5

Artem-Sotnikov:
Silkscreen for these two components have been flipped (this is 100% guaranteed to be assembled in this state) 
image

Artem-Sotnikov:
Good area for small silkscreen logo + board name 
image

rdakhil:
Jack:
1. Done
2. Removed them
3. Done
4. Done, extended PCB so it is flush with the connectors (https://waterloorocketry.slack.com/archives/C07MX0QDS/p1674619268318539)
5. Fixed
6. Fixed, whoops
7. Fixed
8. Fixed for repairability 

Artem:
1. Fixed
2. Put logo + board name on other side of JST connectors

BluCodeGH:
The programming connector now interferes with the jackscrew that will go on the bottom of the CAN connector, it's also _real_ close to that boost converter pin.

eliaskountouris:
Noticed R15 has daisy chaining between pins 3 & 4, is there a layout reason for that?
image

Otherwise (aside from Jack's comment) LGTM

BluCodeGH (changes requested):
Sexy board!

1. I'm not sure exactly what standoffs artem sourced but you probably want those holes to be plated. Also R3, R9, C13 and R16 are all probably going to interfere with the standoff so you should bump them back a bit.
2. The screw head on the back of the CAN connector is still going to interfere with the programming connector. Add like 1mm of clearance at least.
3. Silkscreen missing your name, the revision (R2 I believe) and the date.

rdakhil:
1. DM'd Artem about the standoffs, not sure if this will get fixed in time for the PCB order today. Fixed the components that were too close to the standoffs
2. Done
3. Done

BluCodeGH (approved):
LGTM, although tbh I'd just make those holes plated. Not really a scenario where we want them non-plated.

Artem-Sotnikov:
Why do we want to have them plated

Artem-Sotnikov:
Ok I spoke with Jack and we agree that they should be plated, since otherwise we'd be running traces with no protection underneath mechanical elements which can tighten and bite into the board. We need this to prevent damage to the board - by protecting against tightening damage (even though these screws would likely not be tightened very often).

BluCodeGH:
Looks good.

PR #54, opened 2023-03-13
Add harwin connector 3d models to repo

Rio6:
Add harwin connector step models to the repo so we can add it when exporting cad models

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/54)


BluCodeGH:
Can we make a new folder in lib/ for 3d models? Otherwise LGTM.

BluCodeGH (approved):

PR #58, opened 2023-04-06
On pad charging

jasonskells:
Respin is finished and here's a list of changes that were made:

- Moved the 13V CSA to the VCC rail so we can see the current when the battery is connected and the 24V is not.
- Moved the 13V CSA in the layout so that it isn't directly under the inductor.
- Added a differential filter to both of the current sense amplifiers. (Section 3 page 5 from this application note: https://www.st.com/resource/en/application_note/an4304-how-to-filter-the-input-of-a-highside-current-sensing-stmicroelectronics.pdf)
- Fixed a pin 1 silkscreen issue with the PMOS FET: SI4143DY-T1-GE3. This involved changing the footprint to change the gate to pin 2 and the three source pins to pin 1 and then changing the schematic symbol to reflect that.

Not sure of the revision standard and I couldn't find anything on the confluence so I just put a "rev. 2" on the silkscreen under the board title, let me know if this should be changed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/58)


BluCodeGH (changes requested):
Schematic and PCB LGTM. However, it looks like you made your changes on the same branch as you used for the initial PR and that has confused git, so you should move them to a new branch based off of master.

Also, please remove any outdated gerbers (that's what git history is for) and all gerber zips, so there is a single `gerbers/` folder with up-to-date gerbers and drill files.

Lastly make sure you aren't committing any changes outside of the `charging_on_the_pad` folder, right now this PR has changes to other boards' files.

PR #60, opened 2023-04-13
Charging Board Rev 2 (good PR)

jasonskells:
The merge conflicts have been fixed in this PR, the changes made in this revision are:

Respin is finished and here's a list of changes that were made:

- Moved the 13V CSA to the VCC rail so we can see the current when the battery is connected and the 24V is not.
- Moved the 13V CSA in the layout so that it isn't directly under the inductor.
- Added a differential filter to both of the current sense amplifiers. (Section 3 page 5 from this application note: https://www.st.com/resource/en/application_note/an4304-how-to-filter-the-input-of-a-highside-current-sensing-stmicroelectronics.pdf)
- Fixed a pin 1 silkscreen issue with the PMOS FET: SI4143DY-T1-GE3. This involved changing the footprint to change the gate to pin 2 and the three source pins to pin 1 and then changing the schematic symbol to reflect that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/60)


BluCodeGH (approved):

JekkuChen (approved):
Everything looks good

rdakhil (approved):

PR #63, opened 2023-09-14
add JUST gecko footprint to lib

Joe-Joe-Joe-Joe:
Old PR had all my processor board commits AHHHHHHHHHH

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/63)


BluCodeGH (changes requested):
Looks good, a few thoughts:
- Part number should be G125-MG10605M4P, the 06 refers to the number of pins which definitely changes the footprint and the last P I'm including since having an X there is ambiguous with a part number that just ends in X imo.
- The y component of this dimension should be 6.2mm according to [this drawing](https://cdn.harwin.com/pdfs/G125-MH1XX05M4P.pdf) 
image
- Note that this dimension needs to stay at 7.7mm when you adjust the above
image
- We should make the mounting bolt holes plated. The bolts harwin supplies have a 3.8mm head diameter so we should make both the plating and the back courtyard at least that large (courtyard maybe a bit bigger, 4mm?).
- Right now your back courtyard doesn't include the pins. Not sure if thats an issue.
- I think a silkscreen drawing showing the outline of the connector will be more useful, something like below. The drawing I linked above has all the dimensions you need. 
image

Joe-Joe-Joe-Joe:
I don't think adding a back courtyard area for pins is necessary - KiCAD will enforce space around them as per normal via rules.

BluCodeGH (changes requested):
Sorry for being picky, but we're going to be using this footprint a ton and in places where we want things to be compact.
- Part number is M4P not MP4
- IMO its important for the silkscreen to accurately show the amount of occupied board space. IE it should go right up to the bounds of the current keep out zone on the bottom and sides
- Similarly I think showing the back corners of the top mounting bolt attachments is useful on the silkscreen, like I did in the image in the older review, especially if your front courtyard extends further back than the connector really does. Alternatively, you could make the front courtyard not fully cover the mounting holes and be representative of the actual occupied space.
- The mounting holes should be plated. You do this by setting the pad type to "Through hole" but not assigning a pad number.

Joe-Joe-Joe-Joe:
I went with NPTH originally cause that's how our Datamate footprints are done. I agree it makes sense to change for not like morning the PCB surface.

Do you know how KiCAD deals with them differently though? Like at least on mine when I put an NPTH with a larger annular ring, it still shows up with what looks like plating around the hole.

Joe-Joe-Joe-Joe:
Anyway, I think putting both the courtyard and silkscreen around the outer bounds of the connector on top makes the most sense. I think that gives us enough clearance for the pins on the top side (cause the mounting blocks extend a good bit farther back.

BluCodeGH:
Oh also the front keepout does actually need to also include the plated holes, I just tested and I could overlap an smd pad with the plating. Kicad has a nice arc tool you should be able to use.

BluCodeGH:
Yep! LGTM but please give it a test yourself and make sure it behaves as expected.

Joe-Joe-Joe-Joe:
Soooo this is allowed by default, I suppose since this connector doesn't have a back courtyard. I am not sure how best to fix this, because I think it is fine to have surface mount components on the opposite side, just not THTs. So I don't want to expand the back courtyard.
image

Joe-Joe-Joe-Joe:
These guidelines from KiCAD, which seem to be straight from IPC, specify at least 0.5mm courtyard clearance around the connector. 
https://klc.kicad.org/footprint/f5/f5.3/

This is what you get
image

DomenicHampson (approved):

JasonBrave:
For the pin hole diameter is 0.55 typical, do we have to account for tolerance +- 0.05, and make diameter 0.60?
Otherwise LGTM

Joe-Joe-Joe-Joe:
according to this
https://electronics.stackexchange.com/questions/85292/what-pad-hole-drill-size-is-appropriate-for-a-given-through-hole-lead-diameter
looks like yes. I have made the relevant change

JasonBrave (approved):
LGTM

PR #64, opened 2023-09-15
Processor board schematic

Joe-Joe-Joe-Joe:
I originally speedran the PCB so I'd have something for my workterm report. I plan to redo it from scratch. Don't mind it for now

I am using the following spec for the programming connector, assuming we are in fact purchasing the [STLINK-V3MINIE](https://www.digikey.ca/en/products/detail/stmicroelectronics/STLINK-V3MINIE/16284301). I added some notes where I wasn't exactly sure on pinout:
image

As soon as the PR on the gecko footprint is approved (please approve it!) I will swap the rocketcan connector to that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/64)


Rio6:
Do we care about these errors?

image

Also why are you putting +5V and GND upside down like this XD

image

LGTM

Joe-Joe-Joe-Joe:
PCB is getting redone from scratch so don't mind those errors for now

BluCodeGH (changes requested):
Looks pretty good! The usual bunch of small things:

1. Don't commit your lock files (the ones that start with `~`)
2. What is your expected max current draw? You have a 750 mA PTC but R1 with the op amp limits your sensing to about 200 mA.
3. Why are CANH and CANL global labels rather than netlabels?
4. Is that an approved (tm) 3V3 LDO? We should use the one that other boards use since consistency is good we know some LDOs have issues with RF interference.
5. I'd recommend connecting all of the SD card pins to the MCU's dedicated SD card interface. Worst case you can fall back to SPI (CHECK THIS I'M GUESSING) but using the more advanced protocol will mean faster writes = more CPU time for other stuff.
6. VCAP is the core voltage (like 1.1V or something) but should _not_ be connected to 3V3
7. According to AN5419 your VCAP capacitors should be 2.2u each
8. VDDA should have a ferrite bead between it and 3V3
9. Pls uninvert all your decoupling caps. Netlabels are your friend.
10. The BOOT0 pin should either be a solder jumper or no jumper at all, looks like its meant for selecting between partitions in an A/B update scenario.
11. I'm a big SWD simp and I think you should use it over JTAG, from what I understand its better for dev/debugging whereas JTAG is more intended for testing. Anyways if the pinouts are compatible I guess use both?
12. I'd connect GNDDETECT to GND through a solder jumper that you don't solder (leave NC) by default.
13. Have you checked that all the pins you have assigned work with the peripherals you need? Like eg is UART available over PB8/PB8 and CAN over PA11/PA12?

Joe-Joe-Joe-Joe:
image
oh boy

Joe-Joe-Joe-Joe:
Absolute max current was calculated at ~0.93A, so I bumped the whole power path to 1A.

Joe-Joe-Joe-Joe:
Silkscreen overlaps, mounting hole alignment, edge cuts, extra silkscreen labels still TBD

DomenicHampson (commented):
- Use a pull-up resistor for the reset button. More reliable than a cap.
- The datasheet for the LDO recommends 33uF cap on the output why do you have 100uF?
- TP for CAN lines would be nice.

Joe-Joe-Joe-Joe:
I know for a fact the NRST doesn't have a pullup intentionally
I am going to assume I did that cause this
image

Joe-Joe-Joe-Joe:
re: LDO bulk cap
I am preparing for relatively high transient loads. I would not be opposed to pairing it with a smaller cap (with better ESR) for even better stability.
image

Joe-Joe-Joe-Joe:
re: CAN lines TP
My thinking is that they are already broken out through the Gecko. I would be worried about noise generation from the CANH/CANL traces, which is why they are as short as possible. There isn't really a place to put TPs that wouldn't require me to run the traces out much longer than they are now

I can see your point though... plenty of times where we had a harness failure and misdiagnosed as a CAN failure. I am hoping the new geckos and not being bad at harnessing will fix these issues to some extent.

But you did remind me I forgot something very important, which is GND TPs

Joe-Joe-Joe-Joe:
Backup LDOs (match LM3940 footprint/pinout):

https://www.digikey.ca/en/products/detail/diodes-incorporated/AP2114HA-3-3TRG1/4770627
https://www.digikey.ca/en/products/detail/microchip-technology/MIC39100-3-3WS/771641
https://www.digikey.ca/en/products/detail/diodes-incorporated/AP7361EA-33E-13/16186483

BluCodeGH (changes requested):
6/7 from my old review: Yes that diagram looks right, notice that VCAP = VCORE which is powered by the internal V reg ldo, so that should _definitely not_ be connected to 3V3.

Sorry for the late review this time around:

1. The pinout of the gecko is wrong, it should match the physical layout of the datamates which use a different pin numbering scheme.
2. Call out the part number of SW1.
3. The bead you sourced for FB1 is 0603, also C20 is 0603?
4. In the getting started guide it calls out pin PB9 as "toggles during SD card boot" in the SD connection diagram, maybe move uart to a different set of pins?
5. PC13 - PC15 seem special looking at the pin table, I'd use a more normal pin for the interrupt.
6. 470R is pretty large for LED resistors, also remember blue and white LEDs have like 3.2-3.3V drop.
7. Consider using a female programming connector so its less likely to short during flight, and using male/male header pins to connect to the cable.
8. It would be _really_ nice if the board followed our mounting hole position standard. As a remind thats 1.5in center to center on one edge and a multiple of 0.5in center to center in the other (typically 1 or 1.5)
9. Also right now your mounting holes arent aligned at all.
10. Please use the 3.2MM M3 mounting holes.
11. Rounded corners please :heart: 
12. Does this really need to be a 4 layer board?
13. I'd swap the programming header and the sd card so that the tall stuff is all on one side of the pcb.
14. The getting started guide calls out impedance matching the SD card lines, and that they should all have the same number of vias (ideally zero). Given those are high speed signals they should get routed first.
15. Its nice to label testpoints with their value on the silkscreen, you can do it by editing the  TP settings and changing the layer of the value from F.Fab to F.Silkscreen.

Joe-Joe-Joe-Joe:
Does that mean that the Gecko footprint passed review with the wrong pin numberings?

BluCodeGH:
No, just the old datamates had a funky pinout

JasonBrave:
Why is the JRCLK on programming connector is left unconnected?

Joe-Joe-Joe-Joe:
JRCLK is the return clock signal, which we don't care about. Per this, I've connected it to GND
image

Joe-Joe-Joe-Joe:
re: jack

1. fixed
2. Changed it to whats on the tutorial board, iirc that's pretty standard
3. yes.
4. This shouldn't be the case if the pin isn't multiplexed to the SDMMC peripheral, at least thats how i interpret it
5. fixed
6. I copied minisensor.
7. Done. We really need to remember those male-male pins.
8/9/10/11. Fixed.
12. Yes. Ostensibly because the SD card data lines, mostly because it makes my life easier for routing
13. done
14. lengths are matched and vias are removed. Only potentially sussy thing is trace impedence. I will look into it a bit more before we order
15. I'm doing this last, cause updating from schematic means I have to do it all over again.

BluCodeGH (changes requested):
Looking good! Re my last review:

3. There are no 33uF 0603 capacitors on digikey. Also the footprint for FB1 is currently 0805, but the bead is 0603. Either find a bigger bead (probably preferable?) or fix the footprint?

And new this review:

16. The footprint for SW1 is wrong if you're switching to the standard tht one
17. Probably better to move the GND vias on pins 10 and 99 to below the IC so they are closer to the GNDs on the nearby caps.
18. Remember to update the silkscreen date
19. Pls double check the JLCPCB allowable clearance between vias and the PCB edge for those ones just below the MCU.
20. Also I'm not gonna make you redo it, but if you wanted you could totally put the SD card under the IMU and get this down to 1.5in^2 (which would be really nice since then it would match charging and mini sensor...)

BluCodeGH:
As I understand it, the decoupling capacitors are there to accommodate momentary power spikes going to an IC. So its about preventing that IC from causing noise on the wider rail. So you want the capacitor to be as close as possible to 3V3 _and_ GND so that if there is a power surge the capacitor can easily drain itself to provide.

BluCodeGH (approved):
1. For the third time, **do not connect VCAP to 3V3**.
2. The D0 and D3 traces aren't length matched to the rest.
3. Remember to update test point silkscreens.

QuantumManiac (approved):
I was unable to mention this as I wasn't around during the schematic design phase of this board, but it might be good to write up a little blurb about this board and what it does in the title block. Optimally we should be doing this for all our boards but I think it would be good for this in particular because "processor" is pretty vague for someone having a first look at the boards on this repo.

Otherwise, LGTM

Joe-Joe-Joe-Joe:
What we should _really_ be doing, IMO, is updating the README so it isn't 5 years out of date. That is waaay more effective than a description inside the schematic.

PR #68, opened 2023-11-02
vibration board 1st iteration

IshmanM:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/68)


Joe-Joe-Joe-Joe:
1. Use a net label everywhere you currently have a global label
2. For the SD card, you can and should add separate 3.3V power ports for VDD and the pullup supplies. Also, is there a reason you are pulling up all the data lines? I am pretty confident you shouldn't be pulling up the _unused_ data lines. This is the SD card schematic on Logger board, for reference:
image
3.  CAN termination resistors are wired directly into the bus harness (since there are tons of boards and you need only 1 at each physical end of the bus). We do not add them to individual boards.
4.  You currently have no way to program your microcontroller. Your PIC also has no external oscillator (there should be a 12 MHz external crystal oscillator, see any other PIC18 board in canhw for reference).
5. Why do you have a series resistor connected to MCLR???
6. Please google what a "DC/DC converter" is. You cannot use a voltage divider to step down a voltage supply
image
7. Screw terminals are acceptable for testing and/or ground systems, but for all flight hardware connections we use special locking connectors. Old boards use Harwin Datamates, but we are moving all new boards to Harwin Geckos. For the schematic, you can just use an 01x06 pin header. See processor board (under `canhw/processor`) for the correct pinout on the symbol. Our CAN lines and power distribution are all over the harness. How the payload will connect to the main rocket harness is a different conversation. The harness has a regulated 5V supply, which is what you will want to use as your external power source.
8. D1 will not turn on. The voltage drop of a red LED (the lowest) is minimum 1.7V. That corresponds to about 1/3 of a milliamp running through your diode with a 10k resistor. typical LED operating currents are in the tens of milliamps
9. `canhw/lib` contains a symbol library full of commonly used symbols on the team (PIC18, MCP2562, the SD card I think?). You should add that to your project. Because it is in a parent path to your project, you need to do some directory navigation so that the path will still be valid for anyone who clones the repo. This is how you do it:
image
Best practice is to eventually put your project-specific symbols into the main symbol library, but at this point that's probably unnecessary (and will probably screw up your merge). But you can and should clean it up - extract the symbols from the downloaded stuff, put it in a top level symbol library which you seem to have, and then delete the downloaded files
10. For the accelerometer, I will reiterate that you can and should use multiple power ports to clean up the schematic. On the board side, the team doesn't really have the capability to reliably solder LGA or QFN components. Looking on digikey i can see that you don't have a lot of options. This is why on other boards we simply use breakout boards with the chip already mounted, although it comes at the cost of compactness. I have done a QFN chip once for a board and it was a pretty awful time.

IshmanM:
Reply to @Joe-Joe-Joe-Joe :

Hi Joe, thanks for the feedback. I made a new commit with the following details:

1. Done changing labels.
2. Done separating 3.3V sources. I referred to a few sources when deciding to add the pullups, (mainly intended to prevent floating values, recommended regardless of communication protocol): 
https://www.reddit.com/r/AskElectronics/comments/ek1da3/how_to_properly_make_an_microsd_card_spi_circuit/, https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/sd_pullup_requirements.html#:~:text=SD%20cards%20and%20SDIO%20devices,from%20entering%20a%20wrong%20state, etc...
3. Resistor removed.
4. Oscillator and programming connector added.
5. Datasheet recommends it: 
image
6. Converter added.
7. Terminals changed.
8. Resistance lowered.
9. Library usage updated.
10. Can we talk more about the breakout boards?

BluCodeGH (changes requested):
Looking good! Some things I spotted:
1. The programming header should be 5 pins.
1. You're currently driving that power LED with (5 - 2) / 160 = 18.75 mA, which is a bit too much. Joe was a bit hyperbolic, you should size flight LEDs for 5-10 mA.
1. Also rather than a power LED, it's generally better to have all of the LEDs controlled by the PIC. I see you have some unused pins so lets use those.
1. You should specify that Y1 is 12 MHz
1. The datasheet specifies that the series MCLR resistor should be at most 470 ohms, not kohms :stuck-out-tongue:. That being said, I'd stick with the standard single pullup that we know works and saves on part count. Take a look at another pic18 board to see exactly what we usually do.
1. I'd rename PIC_CANTX0 to PIC_CANTX for consistency with PIC_CANRX.
1. Most flight boards wire CAN_TX and CAN_RX to pins C0 and C1 (but I don't know which is which, look at a different schematic to check).
1. You should also double check that all of our pin assignments on the PIC are mappable to the correct peripherals. Read the PPS (peripheral pin select) section of the PIC18 datasheet.
1. I'd rename the INT0 and INT1 labels to specify which device the interrupt corresponds to.
1. This is a bit of a nit but you should avoid a 4-way intersection in schematics since its hard to differentiate between wires crossing (eg with your ground connection on the MCU. Instead, I've give the LEDs each their own ground.
1. You should specify the current rating of the polyfuse. Most boards of ours draw 10-50 mA, so figure out the expected max current of the vibration sensor and add that. We stock 350 mA and 750 mA PTCs, so I'd guess the 350 will be suitable here.
1. Also double check the max current calculation on that sense resistor. Remember that the ADC max on the PIC is equal to its input voltage and the INA180 has a gain of 100.
1. You can remove the auxilliary battery connection and voltage sensing and diodes, that's a holdover from a GPS-specific thing.
1. You should use the same microsd card symbol and footprint as logger board, since we know it is very robust. Notably I beleive that lacks the detect pin.
1. Since we will be driving CS, CMD and CLK from the PIC to the SD card they don't need pullups, we will just drive them high when they need to be high. The data lines should have pullups.
1. On logger board we use a 4.7u decoupling cap next to the SD card, not sure if that's necessary but we know it works so I'd probably stick with that.
1. Tie the standby pin of the can transciever to ground like on other boards - we don't use it.
1. Remove all of the kicad autosave/lock files from git pls. I'm not sure why they got included in the first place, they should have been .gitignored.

I'll leave you and Joe to figure out what you're doing in terms of accelerometer(s).

IshmanM:
Reply to @BluCodeGH 

Thanks for the feedback. My updates are below. I may need some advice with items 4, 11, and 12.

1. Joe-Joe-Joe-Joe suggested the 6 pin header (Harwin Gecko)
2. Moved LED to control by PIC 3.3V output -> kept 160ohm resistor
3. LED now under PIC control

4. Now specified, but is this clock speed enough?

5. Fixed the resistance value to 470ohm. Will keep it for now, just for safety because of the capacitor.
6. Renamed.
7. Moved signals to C0 and C1.
8. I referred around page 6 and 265 of https://www.mouser.ca/datasheet/2/268/40001943A-1222654.pdf. I think it should be good unless I made a mistake.
9. Renamed.
10. Separate grounds added.

11. I need help with updating the calculation. 
12.  I need help with updating the calculation. 

13. Removed them. I just kept the 1 diode from 5V for safety. Should I remove it aswell?

14. Symbol updated
15. Removed the pullups on CS, CMD, and CLK.
16. Value changed to 4.7uF.
17. Grounded.
18. Added a .gitignore file. I don’t know if it will work/remove the stuff already pushed to the repo?

BluCodeGH (changes requested):
Looking better, some follow-ups and some new points:
1\. There is a difference between the CAN connector (which should be the 6 pin gecko) and the programming connector which is a 5 pin smd female dupont. Please look at another board to get the correct footprint.

5\. I suggest removing both the capacitor and series MCLR resistor. All our other boards work fine and we dont want to deal with jumpers on flight boards, especially around MCLR.

10\. 4-way intersection on VDD of U6. To make it clear what's going on you should use a second +3V3 symbol so that the resistor is placed vertically above the wire.

12\. Your MCU is on 3.3V not 5V. Otherwise your calculation is correct.

13\. You should remove the diode too. Everything is harnessed so reverse polarity protection isn't really a concern.

18\. Looks like you didnt edit the gitignore, so this point is unaddressed. To remove already committed stuff run `git rm --cached /path/to/file`.

19. Please remove the footprint_downloads and vibBoardLib folders. If you need to get footprints they should all be placed in the top-level canhw/lib/canhw_footprints.pretty, and please only ever include the final kicad_sym and kicad_mod files (no zips or html like you have now). Keeping the git repo clean is extremely important since all of the projects share it.
20. I'm not sure why but most of our boards connect LEDs to +3V3 and have the ground end connected to the PIC. Doesn't hurt to be consistent with that thought.
21. We should probably connect the BYP cap to the 3v3 ldo like the datasheet recommends, apparently reduces noise which we definitely care about.

IshmanM:
Reply to @BluCodeGH :

Hi, thanks for the tips. I have some new updates:

Completed steps:
1. Connector replaced with 5 pin one.
5. cap/resistor/jumper on MCLR pin removed.
10. 4-way intersection removed. (The whole accelerometer is replaced, actually)
13. Diode on voltage regulator removed.
18. Just ran the command. I think it should be fixed now.
19. Moved footprint to the top library and deleted vibBoard Library.
20. LED wiring updated.

Maybe completed steps:
12. I replaced the 5V with 3.3V on the Vs pin of op amp. I think this should avoid damaging the PIC now. Is this good?
21. Added capacitor between Vdd and Vss of PIC. Is this the capacitor you were talking about?

Also, I used the AIS3624DQTR to replace the old accelerometer. It is QFN. This accelerometer is connected with I2C as opposed to SPI to save PIC SPI bandwidth for the MicroSD.

BluCodeGH (changes requested):
Per last review:

21. See the datasheet for the 3.3V LDO. It recommends adding a capacitor to the 'BYP' pin (labeled BP on our symbol). Adding the capactor to the pic is a good thought too though.

And going over the new changes:
1. I dont think explicitly differentiating a power LED vs debug LEDs is necessary / useful. Typically we use one of the LEDs as a 'heartbeat' to show both that the board has power but also that the PIC is alive and running code. So I'd make one of the red LEDs green instead and align all of the LEDs horizontally + vertically.
1. I think its clearer to connect both of the PIC VSS pins to the nearby ground and have the .1u cap have its own GND symbol. You don't need to explicitely connect the grounds with a wire.
1. You have duplicated PWR_FLAGs on +5V, one by the CAN connector and one by the LDO.
1. May as well connect the second accelerometer interrupt pin given we have free pins on the PIC, no?
1. It looks like the entire schematic is very slightly off of the correct 0.05in grid? You can fix this by selecting everything, right clicking, and choosing 'align elements to grid'.
1. The current sense resistor should probably be 0805. Also please finish assigning footprints.
2. Do you have changes to the lib/ folder you meant to include? I dont see a symbol defined for the accelerometer anywhere.

Otherwise LGTM, I think you should be good to proceed with layout once the above is addressed without another review.

JasonBrave (approved):

PR #70, opened 2023-11-06
SRAD Camera

BluCodeGH:
As promised at Camp Screep Creep, a camera board using the OV5640 module and an STM32H750. This has been designed for 1080p@30fps but should also support 720p@60fps and 320x240@120fps.

The primary goal is to stream footage to the SD card over DMA, but if we have the RAM for it we may also be able to periodically grab frames and send them over CAN. I've also added USB connector which could potentially be used for grabbing data, but with the primary goal of letting these boards be powered from a USB battery bank for ground tests like static fires.

Design doc that was mostly used for camera and MCU selection: https://docs.google.com/document/d/1vMnUonLazUQGg2DBcX9jqHRnfP3dcwS-uDcvcPPE_Zk/edit?usp=sharing

MCU documentation page where you can get the datasheet and relevant app notes: https://www.st.com/en/microcontrollers-microprocessors/stm32h750-value-line/documentation.html

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/70)


DomenicHampson (changes requested):
Looking nice Jack! Just a few comments/questions...

1. Specify the gain of the op amp

2. Is there a reason why you have the pull ups on the sd card data lines/are not following the circuit from logger board?

3. Shouldn’t the vcap capacitors be 100nF since we are bypassing the internal voltage regulator.
image

4. I couldn't find a pin out for the camera connector that matches yours. Can you share it with me?

DomenicHampson (approved):

JasonBrave (approved):
I think an UART connector can be added, otherwise LGTM

JasonBrave (approved):

DomenicHampson (approved):
lgtm

Joe-Joe-Joe-Joe (approved):
LGTM

PR #71, opened 2023-11-09
Updated usb debug schematic for review

miles-milne:
Created a new kicad project for the CAN usb debug update.
Contains schematic update.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/71)


DomenicHampson (changes requested):
Schematic looks good! Just a couple comments…

For the usb symbol, I would just use the standard kicad usb 2.0 type b symbol it looks cleaner

Same with the fuse, just use the standard symbol wired in series with the 5V line. Also specify the fuse value (did you end up sourcing a fuse?)

You guys accidentally pasted the entire schematic twice so there is double of everything. Make sure to remove all those duplicates. A good rule of thumb is to always use the electrical rules checker, it will catch these kinds of things.

Include the canhw symbol library and update the path in “manage symbol libraries” for the canhw and usbdbg-rescue lib so that the path works for anyone that clones it. Here is how that is done in the old usb debug. Also make a copy of the usbdbg lib inside your folder.
file_usbdbg

JasonBrave (changes requested):
1. Please give the sheet a name
2. Can we replace jumpers with DIP switches?(@BluCodeGH any suggestions?)

BluCodeGH:
DIP switches can be a pain if they're too small, but I don't feel strongly either way (switches vs headers).

BluCodeGH (changes requested):
The reason we use git is that it tracks history for us automatically. So please remove the usbdbg-old folder and rename all of your -v2 files to just plain usbdbg.

Also you should not copy canhw.lib or have any other kicad_mod / kicad_sym files, instead you can reference the parent canhw.lib (see how other boards do it with `${KIPRJMOD/../lib}`) and you should add anything new to that same lib.

BluCodeGH (changes requested):
Looking good, but I'm also gonna use this as an opportunity to be nitpicky about the old schematic so we can bring it up to modern standards.
1. Please remove the usbdbg-rescue.kicad_sym from lib/, all symbols should be correctly referenced from kicad's default libraries or the canhw lib, not rescued. I think you'll need to manually delete the file and then go in kicad and manually check all the footprints to make sure they are up-to-date.
1. Please use the R_US symbol (the squiggly one) for resistors.
1. Resistors between 1 and 999 ohms should have an "R" suffix, eg "330R" for the LEDs.
1. Specify the LED colors as the value, I recommend red and green.
1. I'd use a separate GND symbol for the PIC16 VSS pin and the vusb3v3 capacitor, it will let you make that capacitor a bit less cramped.
1. You can remove the text "Fuse: " from the F1 value, we know its a fuse since its called F1.
1. Generally I like to give two grid ticks of space between a pin and a wire making a turn, eg with the USB connector or the vusb3v3 capactor on the PIC.
1. More specifically with the USB connector though, I think its cleaner to have the fuse be horizontal and then only the +5V symbol sticks up, like this:
image
1. I'd switch the BUS_CONN to be 4 pin, that will definitely help distinguish it from the programming connector.
1. While we're here, it might be nice to break out the CAN and power lines to screw terminals too. I'd probably grab a 4 position 5.08mm screw terminal and throw the same 4 pins from BUS_CONN there too.

DomenicHampson (changes requested):
Looking good almost there. Just one thing please add the USB_DBG and Housings_SOIC footprint libraries. Thanks!

DomenicHampson (changes requested):
Thanks for getting started on this Miles.

Before you started with routing, you should always have the board outline drawn. Refer to the kicad tutorial to find out how to do this. Once you have your outline, re-arrange the components to fit on the board (especially that USB connector because where it currently is would make the board way larger than it has to be). Don't get started on the routing until we've reviewed the layout.

BluCodeGH (changes requested):
Schematic LGTM, nice work!

1. You shouldn't need to add the SOIC or PinHeader footprints to canhw. The schematic was just old so they names weren't being found, you should instead use the updated default footprints (just find the ones with super similar names). We don't want the main canhw_footprints being filled up with duplicates of the built-in ones.
1. We should make sure there are 400mA fuses that fit in that holder. If not we might need to change to the glass cylinder style of fuseholder.
1. Overall this layout looks good, but it also looks very spacious. I think we can get this board significantly smaller if you use the fact that PCBs have two sides and we can overlap components :). Also it would be best for the 4-pin connector and the USB connector to be on opposite sides.
1. Also looks like you've accidentally drawn your board outline on the F.Cu layer, it should be on Edge.Cuts.

Joe-Joe-Joe-Joe:
1. what does "12V" mean here
image
2. try rerunning schematic annotation, not sure why the MCP is like this
image
3. You should calculate the absolute max draw on the 5V line from the board itself, and then make a note somewhere of how much current is available at J3, so people don't aren't accidentally murdering it

BluCodeGH (changes requested):
Layout is looking much better! Some thoughts:
1. Please leave PinHeader_5x2.54x... in canhw/lib, that should be your programming connector and most other boards use it as theirs.
1. See point 2 on my last review. Pretty sure 400mA fuses of that size don't exist.
1. Please add rounded corners, it makes boards look much nicer :)
1. You added a 4-pin connector to canhw/lib but don't seem to be using it?
1. You should have copper fills on both sides of the board. I recommend GND on one side and +5V on the other, it will make your routing much easier / cleaner.
1. Please add silkscreen labels with the pinout of the CAN connector and what each jumper / testpoint is.
1. Your decoupling capacitors all need to be placed as close as possible to the power pins on the ICs they are decoupling.k

DomenicHampson (changes requested):
Looking great so far. Just a few minor changes...

1. Just move the tx label slightly to the right
image

2. You can remove that "cancon" label. What Jack meant is to label all the pins of the connector (ex: canh, canl, gnd, 5v). You can maybe move the resistor and diode a little to left to make space for the label

3. Add a label for just the MCLR pin

Other than that it looks good to me. Thanks Miles!

DomenicHampson (approved):
Looks good! One last thing, don't forget to add the team logo, board name and your name to the board

JasonBrave (changes requested):
1. Please change silkscreen label beside JP2 to "power", and JP1 to "can term" if it fits
2. Add team logo, board name and your name
3. Otherwise LGTM

JasonBrave (approved):
I suggest you to add Month and Year to the silkscreen otherwise LGTM

PR #72, opened 2023-11-12
Propulsion board

WilsonC1234:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/72)


BluCodeGH (changes requested):
Looks like a good start! A number of things I saw:

1. Please update the gitignore to not include the ~lock file, and also please remove all of your changes under actuator/ and the .DS_Store file.
1. We no longer need the IMU, it's moving to processor board. We should keep the barometer tho.
1. We want to be able to control 3 valves, not just one.
1. We are also going to treat the limit switches as basic sensor inputs, so we can remove explicit support for this.
1. Because of the above we can wire all 3 valves through one 6-pin connector, three + lines and three - lines.
1. You already have an I2C bus you are using to talk to the barometer, so you should wire up the IO expander to that same bus rather than using a second one. With I2C. devices have unique IDs so you can pick which one you want to talk to.
1. The PCF8575 seems to have much more functionality (ie pins) than we need, I'd suggest looking for a simpler one with only the 3-4 required outputs which will save on board space.
1. The barometer max voltage is 3.6V on the datasheet, so you still need the 5V->3.3V LDO.
1. The above also means the barometer's I2C lines should be 3.3V, which means that the PIC and everything connected to it should be 3.3V as well.
1. We will always want to use 12V for the actuator power so we can remove the selection jumpers / external power connector.
1. A 750mA polyfuse on 12V likely will be too small given we want to use it to power the actuators, which each can draw like 300 mA.
1. We can remove the 12V->15V DC/DC, we never used it on mini sensor.
1. Speaking of sensors, all of our sensor circuitry should look like the two non-amplified channels currently do. We never used the complicated opamp circuitry on mini sensor either.
1. We want to support at least 5 sensor channels.
1. We'll need to think about what connectors we want exactly, but for now I'd give all 5 channels the 3-pin version (power, data and ground).
1. It looks like there are a ton of extra footprints the eagle import (?), I'd probably remove the import and if you do need to add footprints add them to the common canhw/lib library.

BluCodeGH (changes requested):
Looking better! Things I saw:

1. Please remove all of your changes under actuator/
1. You should remove the disconnected IMU_INT line
1. The Valve connector has 6 pins. We want each RELAY+ RELAY- pair to go to two unique pins, so filling up all 6. You can remove the limit pins and GND pins.
1. Please make all of your labels UPPER_SNAKE_CASE, no spaces or parentheses allowed. Eg `RELAY1_+` or `VALVE_CONTROL_1`.
1. You need to connect the VALVE_CONTROL_n lines on the parent sheet to the mcu schematic. These are called heirarcical labels, you can google that to learn more about how they work. Same goes for the sensor output lines.
1. We will always want to use 12V for the actuator power so we can remove the Valve Control connector.
1. Why do you have two can connectors? Also the CAN connector is 6 pin, the 6th pin is unused. See other boards.
1. Please rename the _AMP_CAPABLE sensor lines since we removed that functionality.

BluCodeGH (changes requested):
Changes look good. I did a more detailed and nitpicky pass this time around:
1. It looks like some of your wires aren't aligned to the right grid. Kicad symbols all use a 0.1" pin spacing and so you should use a 0.05" grid when making your schematics. You can fix this by selecting everything (ctrl+a), right clicking and selecting "align to grid", then going in and manually making sure things are aligned
1. The primary purpose of a schematic is to communicate your design, so you should make it as pretty as possible. The biggest thing here is spacing out components, eg if I have a go at tidying up the actuator power switch I end up with:
image
I'd try to take a similar approach throughout the schematic, eg with your I2C pull-ups, LEDs, etc.
1. We should use a gecko connector for valve control and CAN connectors, its in the canhw lib.
1. You can probably drop _REGULAR from all your sensor lines since its not telling you anything
1. I'd remove the testpoints around the can transciever, the can lines themselves are broken out and can on the pic18f26k83 is pretty well known at this point.
1. Why is your VBAT label off of the PIC not inline like all the others?
1. Probably easiest to stick to a consistent resistor across your leds. You should size for 10 mA at 2V dropout so I get that they should all be 130R?
1. Please make all of your power symbols point upwards (I'm looking at +3V3 on the io expander)
1. Make sure to add NC flags on NC pins (io expander P3)
1. Why is there a custom footprint for the io expander? According to the datasheet it's just SOIC or VSSOP?
1. It's probably a good idea to pop a .1u decoupling cap on the io expander.
1. We should source an SMD fuse for F1. Also 10A is a bit overkill, each actuator can draw maybe a few hundred mA each. So maybe a 2A fuse would be reasonable?
1. Right now your power path to the actuator is the CAN connector -> a 1A PTC -> the +12V rail -> a 10A fuse -> the Actuator_Power rail. So that 10A fuse is doing nothing since the 1A PTC is going to trip first. You should derive Actuator_Power directly from the CAN connector instead. Note that you should move the +12V current sensing to still include both rails, so probably the way to go is CAN connector -> 2A fuse -> current sensing -> Actuator_Power -> 350mA PTC -> +12V.
1. 1A is quite high for the polyfuse given that 12V rail should only be powering the sensors. I'd go to a 350 mA PTC instead (I pick 350 since its the smallest size we stock).
1. Same thing applies for the +5V PTC, the digital electronics here really shouldnt be drawing much at all but 350 mA is our standard size.
1. The barometric sensor schematic would be a bit cleaner if you connected PS to the CSB pin, and the right CSB pin to VDD, so the left +3V3 can only power the left CSB pin.
1. I'd switch to solder jumpers instead of 0 ohm resistors for sensor voltage selection. Does the same thing but smaller footprint and saves a part
1. Probably wouldn't hurt to add a +5V option for powering the sensors.
1. The layout of your sensor circuitry is a bit all over the place. Remember that since you're using labels they dont all need to be wired to the same place, I'd isntead move the output labels and test points next to each individual unit. Also all the stuff about laying stuff out prettily applies.
1. A neat trick you can do here is because all of the sensors are the same circuit, you can make that circuit on a subsheet, and include multiple copies of that subsheet. So you only end up with one copy of the schematic you have to make changes to.
1. The last thing to do is to figure out what we're doing in terms of sensor connectors. I think the molex micro-lock plus line looks pretty good, so I'd tentatively plan to use those (I ordered some samples so we can see if they are as good as they look before committing). Based on what needs to plug in here we need at least 3 of the sensors to be 3-pin, so making the other two 2-pin (omitting the GND connection, which isn't used for 4-20mA sensors) seems reasonable to me.

BluCodeGH (changes requested):
Almost there, mostly just some footprints to fix now:

1. Please revert the changes in this pr to sensor_mini
1. `RELAY3_-` isn't connected.
1. You should source a surface-mount fuse for F3. You should be able to find one with an 0805 or 1206 footprint, please include the part number as the 'value' in the schematic.
1. You should power the INA180s with 3.3V so they cant overvolt the PIC.
1. With the 100V/V gain of the INA180, you will get 3.3V out on the current sense line (the max) at only 165 mA of current through R3 and R10. We want to be able to sense up to the rating of the fuses, so you should adjust the values of R3 and R10 to allow us to sense 2A and 350 mA respectively. You can look at https://parts.waterloorocketry.com to see what resistor values we already stock and use them if they work.
1. Please use the R_US symbol for the resistors in the analog sensor input sheets.
1. The decoupling capacitor for the PCA9536 should point downwards, as should its GND symbol.
1. You should connect the STBY pin of the MCP2562 directly to its GND pin so you dont have an upside-down GND symbol.
1. You can probably remove the testpoints on CAN_TX and CAN_RX, we have a pretty good idea of how to get CAN working on the PIC18F26K83 at this point.
1. The footprint for U7 cant be found, its trying to pull from a nonexisting 'sensor_mini' library.
1. Please assign footprints to the solder jumpers and test points.

DomenicHampson (changes requested):
Hey Wilson looking good. Almost there just a couple small comments...

1. 20mR for r3 is still too high since 2A will produce a reading of 4V and our mcu can only detect up to 3.3V. We’d like a resistor around 15mR

2. Same for R10 it’s just slightly above 3.3V

3. Use only one gnd symbol for the pic on the vss pins

4. Use 100R resistors for the leds since we have them in stock and at 130R they are only receiving 10mA of current

5. Use the hand soldering footprints for the resistors 

6. Still have the issue with the sensor mini footprints. Message me on slack if you need help fixing this

DomenicHampson:
Just a few small changes then you have my approval...

Re annotate the schematic again but make sure to select reset existing annotations under options.

Let's make R8 90mR

DomenicHampson (approved):

BluCodeGH (approved):
I think you should be good to proceed with layout now. A couple last things from me on the schematic:

1. Typing "FUSE 0805-FF PW 2A0" into digikey yields no results. Are you sure you have the right part number? Also you should set the value field to just the part number and leave a text note to specify the current rating.
1. In terms of R10, I'd probably go a little lower, since PTCs dont fire immediately and its nice to have a bit of breating room to measure higher currents. Maybe something like 62 mR (which I say because we already have some, in 0805)
1. 100R is probably fine for the LEDs but FYI @DomenicHampson  we typically aim for 10 mA as a rough upper bound on LED current.
1. A fully charged lipo is about 12.6V and on the rocket that line can be as high as 13V, so you'll need to tweak your VBAT voltage divider.

DomenicHampson (changes requested):
Hey Wilson looking good so far...

1. Please add mounting holes to the schematic and pcb and tie them to gnd. Refer to other boards for the correct diameter hole size. You will also have to move some components around near the corners to fit them in.

2. The programming connector J1 is facing the wrong way. Please flip it around and align the conn with the edge of the board. Like this:
image

3. We can also decrease the size of this board. Be sure to use the other side of the board for components too. Honestly it might be worth redoing the pcb layout and deleting the current routing. Once you redo the layout let me know and I will get to reviewing it asap. Because we are redoing the layout I won't leave any comments on the routing for this review.

4.  Align all other connectors with the edge of the board. This saves space for components and looks cleaner

5. Place the crystal oscillator closer to it's pins on the mcu. Also make the layout of the crystal and it's capacitors symmetrical about the mcu pins and add some gnd vias around it to help with signal conditioning. You also want the traces to be the same length. Doesn't have to be exactly like this but this is how I did it on an older board
image

6. Make sure capacitors are close to their associated power pins. For example C2 should be close to the 3.3V pin of the mcu
image

7. I would isolate the barometric pressure and temperature sensor from heat generating components on the board.

8. Group the test points together in a corner of the board
 
9. If you can try and group all the test points together in a corner of the board.

10. Also add the rocketry logo, your name and board name after you redo the layout. Refer to other boards to see what this looks like.

Reach out if you need any help addressing these comments. Thanks!

DomenicHampson (changes requested):
Hey Wilson looking better. A couple things...

1. The top side of the PCB is very crammed. I would recommend putting components on the back side to save space (sorry I missed this earlier). This will help create more space for the silkscreen labels since a lot of them are either off the board or under components. All silkscreen labels must be visible and associable with a part.

2. There's something up with these vias around the oscillator. I would delete them and replace them. Make sure you see the GND label.
image

3. Avoid placing vias over pads it can cause problems when soldering. Go through the entire board looking for these.
image
image

For now just focus on moving components to the back side of the pcb to make space. I will give a more thorough review when that is done. 

Thanks for all your hard work Wilson!

Rio6:
All these ground vias are connected to the single layer only - there's no ground layer on the top. You probably want to have another ground pour around the crystal otherwise these ground vias are doing nothing.

圖片

JasonBrave (changes requested):
1. The via beside JP14 might be to close to the jumper
2. You may want to add more vias around the crystal to improve isolation
3. Could the crystal moved a bit closer to the MCU?

PR #73, opened 2023-11-21
charging board respin schematic

Lucifersan:
https://www.ti.com/lit/ds/symlink/lmr51430.pdf <-datasheet for new buck

added new buck, found inductor that maybe works, changed the capacitor/resistor values to hopefully be correct

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/73)


Joe-Joe-Joe-Joe:
Servo connection pls?

Lucifersan:
> Servo connection pls?

servo connector is microcontroller no? we're PRing for the buck part specifically, or should we do everything in one go and come back with a new bigger PR

Joe-Joe-Joe-Joe:
Yes, best practice is to complete the schematic changes to the best of your ability before opening the PR. It makes reviewers' lives much easier

AshTheEngiqueer:
Schematic updated with motor connectors and pins

AshTheEngiqueer:
1. done
2. not super sure what discussion you;re referencing?
3. done
4. Luci and I don't know about this one - won't it be dependent on the motor? Also if the motors are different, should we do a different input/output for each?
5. that makes sense, +13V was me not really understanding the schematic and just picking something that sounded right so thats on me
6. makes sense, assuming the gate
7. Added
8. Already considered, forgot to update note
9. webench says 453k so we're assuming that's fine? not sure though - thoughts?
10. have not looked at yet, will review in the next few days
11. not really looked at yet, that makes sense when I think about it though
12. think this one is fine, looked for smd with same (similar) specs - temp, voltage, operating temp
13. searched for the INA2180 and 4180 on digikey with no luck, moved to filtering for more than 1 channel current sensor with output to CAN bus
14. quick but hasn't been done yet

AshTheEngiqueer:
Still on todo list: check motor outputs to see if voltage divider needed, look at inductors, look at current sensing

AshTheEngiqueer:
Not sure if we can determine whether voltage divider for motor output is needed without knowing what motor is being used? Also a few struggles with changing the INA180 to a INA4180, namely we can't find the symbol in Kicad.

Lucifersan:
I was too braindead at the time to remember to mention this, but I think we can import footprints. I did it once, but I'm not sure I did it correctly. Could probably try that if footprints for INA4180 are available somewhere.

AshTheEngiqueer:
i believe that's everything changed

BluCodeGH (changes requested):
Almost there :sweat_smile: 
1. Looks like you've accidentally committed a bunch of .DS_Store files in addition to changes in remote_arming? Pls fix.
1. The point of the FET on the power to the motor is so that the we (the PIC18) can control power to/from the motor. So you need one pin from the PIC to the FET (because this is high-side switching you probably need a second FET, see the +5V switching circuit), and another from the PIC directly to the motor for PWM.
1. You should add the ability to put a voltage divider between the analog output from the motor and the PIC, just in case we need it.
1. No part number on the INA4180 symbol?
1. You can remove the netlabels from the current sensing subsheet, the nets will use the label from the heirarchical pin.
1. You should move the +13V and +BATT current sensing to before the FETs on the LTC4414. It just helps keep them separate from the ORing logic so its easier to see that they will work correctly. Similarly I'd move the motor current sensing to before the FET.
1. Since the motor control header is mostly an output I think it would be a little more clear if you flip it around to point to the right.

BluCodeGH (changes requested):
Did a more thorough review this time around and caught some more things to fix:
1. You still have a .DS_Store in the root of the respository. Also why are there three new FUSE_BEL footprints when only one is being used?
1. You need to use one P-fet and one N-fet for high-side switching, you can't use the same part twice. I dont actually know how to properly choose FETs to use, you'll need to do some research.
1. You don't need to label 'GROUND' on the motor connector when it is wired right to a GND symbol lol
1. The main sheet is getting pretty crowded, maybe you want to move the motor control stuff to a separate subsheet?
1. You should have R30 as 0 ohm and R31 as DNP in the schematic, and leave a text note explaining it is for an optional divider.
1. You removed the heirarchical labels (yellow) from the current sense sheet and left the netlabels (blue). The heirarchical lables are what show up on the parent sheet, so you need those. The blue netlabels are labels that will locally connect within the same sheet, in most cases they are optional but good practice so that all the wires are labeled. If you have a heirarchical label then that becomes the name of the wire, so you dont need the netlabel.
1. If you're having trouble fitting the +13V current sensing it by the diode OR, might be smarter to put it at the output of the +13V buck just like how +5V sensing is done?
1. What connector are you going to use for the motor?
1. The value for the 5A fuse is still the old one with the big holder, but it looks like you've updated the footprint and datasheet? Please set the value to the part number of the fuse you want to use. Also I'd really suggesting just using an 0805 fuse, there are lots of options on digikey.
1. The 13V buck's inductor is still using the old footprint. Also consider the SRP1038CC-100M instead, it has a much smaller footprint than the one you sourced, and kicad already has the footprint built-in. Second result on digikey filtering for SMD inductors with a current rating of at least 8A, sorting by price.
1. Please fix the path of the footprint library to be relative to `${KIPRJMOD}`, see how its done in other boards.
1. Also you're not including the main canhw symbol library which means the symbols for the PIC18, etc are 'missing'. Again see other boards for how to include it. Also pls remove the "New Library".kicad_sym that it looks was accidentally added?
1. You should update the CAN connector footprint to the new gecko connectors. See processor board for an example.

JasonBrave (approved):

BluCodeGH (changes requested):
Still found a bunch of small things to fix, but I think once these are addressed you can go ahead to layout without another review.
1. Point 1 of last review about FUSE_BEL footprints was not addressed?
1. Point 10 of last review was not addressed? I was talking about the 24V->13V buck. The 12V->5V was fine and shouldn't be changed.
1. I'd suggest moving Q4 to match where it was on the old schematic, whch also matches the datasheet and it a bit clearer with what is happening (current flows down and to the right).
image
1. There is no footprint assigned to the LMR51403, or the +13V or +BATT current sense resistors.
1. Taking a look at C29, 47uF capacitors rated to at least 26V (double expected working voltage) dont exist on digikey in 0805.
1. Give yourself some breathing room around the components in the motor control sheet. Look at how the high-side switching is done on the power sheet, thats much more comfortable to read. Typically I aim for 2 grid spaces of straight wire off of every pin, except when connecting directly to a GND or power symbol.
1. The motor current sense labels aren't connected to the resistor.
1. Its bad practice to have multiple labels on the same wire. Eg your MOTOR_V and MOTOR_V2 implies there is a difference between the two but they are directly connected, so just have a single MOTOR_V label.
1. I'd probably rename MOTOR_OUT_CHIP to MOTOR_OUTPUT_DIV to make it more clear what's going on.
1. Your motor control connector will be easier to follow if you have a single GND symbol just below the connector and allow wires to cross, like this:
image
1. Looks like you're using the wrong symbol for Q8, maybe you need to re-copy from the power sheet?

DomenicHampson (changes requested):
Looking good. Just a few small things I caught

1. Update the footprint for the servo connector to the data mate
2. Specify the gain of the op amp
3. Delete the unnecessary +5V_SW net label in the power sheet

Joe-Joe-Joe-Joe:
Can you add a 0 ohm at the output of the 5V buck? Otherwise we wont be able to power cycle the motor controller copies, since the 5V regulator circuit will keep the MCU on in perpetuity

DomenicHampson (approved):

DomenicHampson (dismissed):
Good start to the PCB layout, here are some recommended changes...

1. You have two test points directly under the holes for the crystal oscillator. I would group all the test points together in one spot

2. In general, try your best to align components for a cleaner look for example these resistors here
image

3. Cap C17 silkscreen label is overlapping with L1. Same with D4. Also the L1 foot print is off the board please move it too. Here are some other silkscreen labels that need to be adjusted: R3, L2, C24, C12, C7, U1, C22

4. Rotate c9 180 degrees. Look at all the components and see if there are others it makes more sense to have in the other orientation.

5. Add a label for the mclr pin on the prog connector.

6. You need to move Q8 up and make sure there’s enough space around the harwin connectors. Those blue circles represent the screw position so right now the connector is covering Q8

7. Place the crystal oscillator near the mcu with the decoupling caps at a symmetrical distance from the crystal… here’s a pic of what I mean
image

8. Make space for the rocketry logo, board name and your names. Behind the programming conn is probably the best spot

BluCodeGH (commented):
Jumping in with a last-minute schematic review, I'll let you clean up the PCB before I review the layout.

1. JP2 should be between +5V and +5V_SW, without the mosfet in the way. The point of it is to allow us to bypass that mosfet for when we want power to flow backwards from +5V_SW to +5V.
1. It would be really nice to still have +5V current sensing when we are drawing power from the bus. If we move JP1 to before the current sense resistor, and then have JP2 connect +5V_SW to before that resistor too, then in either case power flows through the current sense resistor so we can monitor it.
1. We should also add a 350mA PTC in series with JP2 since we dont want a short to be able to take down the bus.
So in essence the above three points are proposing this:
image

1. You should include the "1W" requirement on R23 as part of the value, not a text note. Also do you really need 1W of dissipation? At 2 A that resistor is only dissipating I^2R = 80 mW lol.
1. The text on the power sheet about tuning the RC filters should be in the current sense subsheet, along with the "5V rail CSA" text.
1. Like Joe mentioned in slack, if you need to power a 5A motor, you probably want at least a 6A fuse for F1.
1. You have the wrong footprint for the CAN connector, right now you have the old datamate but it should be the gecko.

Joe-Joe-Joe-Joe:
1. For the oscillator, read section 2.5 of the PIC datasheet and follow the placement guidelines as best you can. You have the space for it. The unshielded inductor sitting beside it worries me. If you want to do an actual via fence in addition to a GND pour there's a whole rabbit hole about it. But right now the four GND vias on the corners of the keepout aren't doing anything special.
image
2. Seems like most of the outputs on U7 go to the opposite side of the PIC. Worth checking if you can swap EN_5V and CURR_MOTOR pins, then you can perhaps move u7 to the other side and not have to run traces so far. Maybe under J2? 
3. I don't have the time to test out a "better" layout on the voltage converters, and perhaps you have read the example layouts and this layout is a compromise for compactness. If you haven't done that, I would try to lay out the converter circuits off to the side and follow the example layouts and guidelines as best you can, and then work towards fitting them into the board confines. I think you can be less generous about spacing a lot of your passives, which will give you more room overall to play with. You can place silkscreen labels in line with your passives, and then butt them up against each other, like so:
image
4. No shame in a 4 layer board, despite what Jack may say. If it comes to that. Always better than making your board bigger, because of our rocket volume constraints.
5. I would prefer obvious silkscreen labels on all the jumpers that say _what_ they jump
6. Double check the frontal clearance on J4. J1 you can see has a pretty big area because it is horizontal. It wont matter much if we plug a shielded Datamate into J4 anyway, but if you have a lot of overhang on the board-side connector it makes it easier to accidentally morb

Rio6:
You should run the design rules checker on the PCB, it is showing a lot of errors right now, for example:

1. This ground via from layer 2-4 is going through the CHG_CURR on layer 3, near C34
圖片

2. Near D4, is 24V and VIN are touching each other
圖片

3. And many other clearance violations. Some of them seem to be fixable by reducing the trace width if they don't need to carry current.

Some other things:
4. Why not connect all the 5V together and save a via?
圖片

5. Looks like you can give some more space between J1 and J3, wires and screw drivers will need to get in there and more space makes it easier.

Side note, I'm not sure why this is used over traces but I like the modern art XD
圖片

Joe-Joe-Joe-Joe:
1. The length tuning might be overkill, and the widths aren't match, but your CANH/CANL should look more like this. I'm confused why you were using a unnecessary via to the front layer. Likewise, what are you doing with CAN Rx here?
image
2. If you want to separate these two pours I imagine you will need more clearance. Double check JLC's spec, but I expect this is less gap than their tolerancing allows
image
3. If you run the oscillator like this you have less perimeter = less fence. Also, are you using 2 layers of vias to get the required spacing? This may or may not be useful, since the left side of the board seems to be not much more than LEDs and a programming connector. But since you're going to have do redo routing without blind vias anyway, something to consider
image
4. These traces seem undersized given this is the high-current path. Regardless I'd just make this a pour
image
5. This trace is _definitely_ undersized
image

miles-milne:
Board looks awesome. Just some nitpicks:

1. Small thing but it looks like there's an extraneous newline by Elizabeth's name (unless they wanted it like that). I would also probably add the team logo near your names too because there is some space there.
image

2. Unconnected part here:
image
Can most likely be resolved by doing something like this:
image

3. Right angle trace here:
image
Believe the team prefers 45deg turns for routing e.g.: 
image

Another one here too:
image

4. Some other examples of sus traces that could be cleaned up a bit

before: Screenshot 2024-03-12 191348
after: image

You can also shorten these ones if you want
image
image


5. After a quick DRC check it stated that some of the pads for the ICs didn't get an appropriate access to the pours. Here's an example of one that (I think) I fxed:
before: image
after:image

6. There were also some other things on the DRC that I wasn't sure how or if they needed to be resolved, so maybe check with an electrical lead about that. 

Other than that the PCB looks good to me.

Joe-Joe-Joe-Joe:
1. See image. You have space to route the caps exactly like that
2. Wherever you have a "thermal relief connection to zone incomplete" consider if it is cleaner to just run a trace directly from pad to via. usually that means you have an isolated pour, and so you end up getting a narrowed trace because it is trying to do thermal relief clearances
image
3. Was the plan to isolate the oscillator ground? If so, you haven't done that on the GND plane. Same on bottom layer
image
image
4. I would label test points with what they are rather than TP designations
5. Tradition is to give your board a "fun" name, as opposed to just "Charging Board 2024". If you want. I feel like you two are extra enough to come up with something

JasonBrave:
You might need a larger via for the +13V line

Joe-Joe-Joe-Joe (approved):

JasonBrave (approved):

Joe-Joe-Joe-Joe (approved):
LGTM

DomenicHampson (changes requested):
Thanks for all the hard work Ash! A few small things then you are good.

1. Are the traces thick enough for the expected current on 5V_SW? Also same concern regarding the via. Let me know what the expected current draw is and your calculations for trace/via size.
image

2. Same concern with the VCC and R_BATT_N vias
image

3. I am seeing some of the comments I left in our messages unaddressed have they been resolved/is there a reason they were not addressed?

4. Also did we decide to just leave this as is?
image


Thanks

Joe-Joe-Joe-Joe:
@DomenicHampson please `git pull`

DomenicHampson:
Oh bruh I thought I did

DomenicHampson (approved):

BluCodeGH (approved):

Joe-Joe-Joe-Joe (dismissed):
Sorry for the late review. Pretty much everything I have is nits about silkscreen and I want the damn boards.

For future reference, the way to do test point silkscreens is to move the actual footprint labels (TPx) off into the ether, well outside of the edge cuts boundary, and then generate extra text objects on the appropriate silkscreen layer. The reason to do this is twofold: one, the DRC won't complain, and two, if you need to update the PCB with changes from schematic, your naming changes won't be overridden. But again, just for reference.

Also you need to generate drill files. Also also, you should rename your inner copper layers to PWR and GND respectively. And then regenerate gerbers. This will make it possible to distinguish which gerber is which when you go to set the layer stackup.

LGTM
image

Joe-Joe-Joe-Joe (approved):

JasonBrave (approved):

DomenicHampson (approved):
Looks good thanks Ash

PR #74, opened 2023-11-29
DAQ CAN Support

KavinSatheeskumar:
- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/74)


DomenicHampson (changes requested):
Looking good thanks Kavin!

1. As you already know, add the resistor on the base of the mosfet

2. I would use the same relay from can support it’s a lot cheaper (G5LE-1A)

3. (As seen in the can support schematic) add a diode between the relay control pins to provide a path for the current to circulate when the coil is de-energized

4. I would use a mosfet as opposed to the bjt you selected. They typically require less current to drive the gate and are just overall better for this application.

5. Mine as well throw in that pull down on the mosfet gate too (as seen in CAN support)

6. I’m pretty sure we don’t need the can termination resistor you can remove that

7. Use a female usb b connector instead so we can plug it into daq via cable

8. Quite a few errors running the electrical rules checker plz run through them

JasonBrave (changes requested):
1. What's this wire beside USB port? You don't need wire over GND, just connect to it
image
2. The file should be renamed to DAQ CAN support to avoid confusion
3. I suggest put micro-controller + CAN and Relay into two separate sheets
4. I suggest connect the CANH and CANL wires using labels instead of running a wire

JasonBrave:
1. I think some of the [comments on the USB Debug respin](https://github.com/waterloo-rocketry/canhw/pull/71#pullrequestreview-1773322019) also applies here
2. Please finish assigning footprints
3. I didn't saw a fuse :eyes:

BluCodeGH (changes requested):
Looks good, just a few small things:
1. Remove `report.txt`?
1. You should add a 350mA PTC (I say 350mA since we stock them and we dont need to power other stuff via 5V, unlike on usb debug) on the VBUS line from the USB connector, dont want to accidentally fry something.
1. Why is the transistor pull-down 210K? A common value for a pulldown would be 10k or 100k, 10k is probably good here?
1. I dont think you have the right footprint for your relay. Tbh might be easier to source a different relay that kicad already has a footprint for lol, there are lots of `Relay_SPST_...`
1. I also dont think your transistor has the right footprint, its a SOT-323 on digikey.
1. You should source a part number for D1. Also flip it around (press 'x' with it selected) so the text isnt overlapping the relay.
1. Its good practice to avoid 4-way intersections of wires on a schematic since its hard to tell whether you meant the wires to cross without touching. I'd have the line to the screw terminal (relay pin 3) have its own +24V symbol since its sorta independent of the voltage you're driving the coil with.

And yeah most of my comments on usb debug apply here too, so take a perusal through that.

DomenicHampson (changes requested):
Looking good a few small changes though...

1. I’d move the usb stuff to the left of the programming connector just to group the connectors and space things out better

2. Let’s put some test points on canh and canl too

3. Renumber the relay so it matches the numbering on the footprint

4. Move gnd to the 4th pin of screw terminal J4

5. Re-annotate the schematic it skipped J2

6. I think the text "Al" under c4 is a mistake?

7. Make sure to add the path to the canhw footprints under footprint libraries

8. Same for the Housings SOIC library

Joe-Joe-Joe-Joe:
1. Pls add datasheet link for your PIC of choice inside the symbol properties
2. Am I high? Where does this board get +24V from?

DomenicHampson (commented):
Joe is right idk how we missed this. Please add a 2 pin screw terminal to the schematic for 24V input

ManavToor:
I have no idea how the 24V input went unnoticed for so long lmao, but it's been added alongside the datasheet for the PIC. I also went ahead and did the layout.

DomenicHampson (changes requested):
Looking good so far.

1. Add the screw terminal for the 24V line. Oh I see you are planning on using the 5th terminal for 24V in. I feel like it would be cleaner and we'd be less likely to wire things incorrectly if we have a separate screw terminal. Make sure that second screw terminal is on the side of the board closest to where the 24V supply is coming from.

2. For the can connecter silk screen label, decrease the font size and realign the labels. One of them is overlapping with the mounting hole. I would also move the screw terminal closer to the edge of the board and center it between the mounting holes.

3. I would use the same mounting hole footprint as seen on other boards for consistency.

4. What is the expected current draw on the 24V line? Are you sure your trace is thick enough the trace is thicker on the old can support board.

5. I just realized you have the wrong footprint for the usb connector. We want a usb b not a micro usb b. The usb b is more durable. Also have the connector going right to the edge of the board. Same with the prog conn.

6. You can remove this gnd via here
image

7. I'd move these vias slightly so they aren't overlapping the thin red line
image
image

Thanks!

ManavToor:
Separated the screw terminals for the 24V traces

I think I'll keep the holes, they look pretty neat and they are standard M3 screws

According to Kavin, current draw is 3A so I calculated a width of 53.83mil

Changed the usb-b-micro to just usb-b

DomenicHampson (changes requested):
Looking better...

1. Please move the usb b connector more inwards. Having it hang off like that can lead to mechanical stress on the connector and pcb. Not any farther out than this 
image

2. Are there dimension requirements on the board? If not adjust the board size or hole placements so that they are closer to the edge
image

3. Have the prog connector going up to the edge of the board (looks cleaner)

4. Nitpicky but can you align these labels a little better
image

5. Instead of all these gnd vias route the pins to each other and have just one via. Same thing in other place on the board
image

6. The datasheet recommends this cap on the TVS diode. Also am I blind I could not find the pin out in the datasheet. Are you sure you have the correct pin out?
image

7. Instead of Vin and Vout, make that two pin screw terminal Vin and GND and then the other screw terminal a 4 pin with Vout there. We want that ground connection plus it cleans up harnessing.
image

8. I also realized there are no testpoints on this board. Add a test point for... Group them up in the same corner of the board

ManavToor:
1-5, 7, 9
Done!

6
Added the cap, pin 1 is labelled in the image you sent

8
Added test points to 5V and 24V. 
Other testpoints were removed here
> 2. We don't need testpoints on CANH and CANL since they go right out to the screw terminal. Honestly I'd remove the existing testpoints too since at this point CAN is very well characterized with these chips.

Artem-Sotnikov:
Looks good, I see no issues that would affect the functionality of the board when in use. I do have a few minor improvements to suggest.

1. The mounting holes are not spaced in an exact rectangle for some reason and have some funny dimensions. I don't see a reason why they can't be spaced as a 2x1 inch rectangle.
2. Since there is only one ground plane, I think the traces on the back should be used more sparingly then they are here. There are a few traces that could be routed creatively on the front without needing to dip into the ground plane, and some of the other traces on the back could be shrunk. Might be worth it to throw on a few ground stitches across these back traces as well. 

One last suggestion would require a schematic change so it's really up to you guys whether or not you want to implement it, but an always-on LED on the output of the relay as well could be quite useful for debugging. Always-on outputs for relay controls have been massively helpful in the design and debugging of the FYDP relay board.

JasonBrave (changes requested):
1. Please double check if USB type-B connector actually fit in the board, or is it intended for the connector go outside the bard boundary.
2. You may want to use a pin header footprint instead of a pin socket footprint for JP2
3. Otherwise LGTM

JasonBrave:
Another reminder for you to add MCLR silkscreen

JasonBrave (approved):

PR #75, opened 2023-12-28
Vibration board PCB layout 1st iteration

IshmanM:


BluCodeGH (changes requested):
Looks like there are still a few schematic changes required, but I did a cursory review of the layout as well:

1. Please don't commit ~.lck lock files.
1. Per the 3V3 ldo datasheet the bypass capacitor should be 10 nF.
1. Points 2, 4 and 5 of my last review weren't addressed?
1. In terms of the layout, your mounting holes aren't quite square. The horizontal distance is > 1.5in.
1. Generally you've got a ton of unused space on the back of the board, you could probably fit this in a 1in x 1.5in hole pattern if you wanted to. Feel free to put more than just passives on the back.
1. In the future I strongly suggest you avoid most of the routing until the layout is approved so that you dont need to redo it if changes are requested. Doing some routing of any high-speed / important traces is fine if you want, but you dont need to worry about the nitty gritty stuff like LEDs or power at the layout stage.
1. Try to keep decoupling capacitors on the same side of the PCB as the ICs they are for. Also remember that decoupling capacitors form a loop that includes the GND connection back to the IC, so you need to make that GND trace as short as possible too.
1. The programming connector should be placed right at the edge of the board, not back slightly like it is now.
1. If you flip the CAN transceiver around to the back of the PCB the routing of CANH and CANL will be much nicer since they wont have to cross.
1. I suggest having one side of the PCB be a +3.3V fill and the other side be a GND fill, that way you dont need a big snakey +3.3V trace.
1. If you can I'd suggest placing the accelerometer right at the exact center of the PCB. That way you can mount the PCB in the center of the rocket and so any spinning will be about the axis of the accelerometer.
1. I'm seeing a lot of daisy-chaining here, that's routing a connection to one component through the pad of another component. The concern is that if a pad rips up, we dont want that to disconnect components down-the-line.
image

IshmanM:
Reply to @BluCodeGH 

Thanks for the review. 

Based on your feedback  I decided to just restart the PCB layout. For now I just made a layout of the key components (excluding the resistors and caps) to get approved. There is no routing yet.

About the schematic itself
- I changed the bypass cap to 10nF
- Missed points 2 and 5 from the schematic review are now addressed
- About missed point 4. It is now somewhat addressed. There are not enough INTx pins on the PIC since those pins are needed for the SPI and I2C buses. As an alternative I just connected the 2nd interrupt pin of the accelerometer to an interrupt-on-change (IOC) pin on the PIC.

BluCodeGH (approved):
The schematic is still off-grid for me. Make sure you right click and have the 0.05in grid selected, then select all and right click -> align elements to grid.

This layout looks much nicer. A couple tweaks I'd make:
1. You can align the CAN connector with the edge of the board by making the silkscreen line go up to the edge, not the courtyard line.
image
1. You should put the crystal as close as possible to the pins it connects to on the PIC.
1. It looks like you're only using the QFN24L_4X4X0P5_STM.kicad_mod, so you can remove the -L and -M variants.

IshmanM:
Reply to @BluCodeGH :

Thanks for the feedback.

I made all the fixes and finished routing the PCB. If there's anything to change just let me know.

IshmanM:
Also, I realized the green and blue resistor values are potentially too high. I guess if that's the case when tested a lower resistor can be soldered

JasonBrave (dismissed):

BluCodeGH (changes requested):
This looks really good! Just a handful of small things to fix:

1. You're going to have a hard time fitting tweezers + a soldering iron around C7/C8 since they are so close to the IMU. I'd recommend giving them a bit more space from both the IMU and each other.
1. The routing for the IOC line looks like you could tidy it up a bit.
1. C3 is also going to be hard to solder, if you rotate it 90 degrees it will work better. Generally you should space out the capacitors a bit further from U1, maybe half a capacitor width of space or so.
1. I'm seeing a fair bit of daisy-chaining here. We want to avoid routing signals _through_ a pad, since if that pad then tears off it breaks the connection. Eg: MCLR pin on PIC, +5V pin on LDO and CAN transciever, +3V3 on can transciver's capacitor, etc.
1. You should try to route CANH and CANL right next to each other, since it's a differential signal. I'd recommend having both of them go above the screw hole on the CAN connector and having two vias right where the +3V3 capacitor is currently. You can shift the capacitor down a little to make room.
1. IMO it would look better if you center each resistor under its LED rather than squeezing them together.
1. You should add a silkscreen marking showing where MCLR is on the programming connector and that it is a 3.3V board. Typically just some text along the side that says "~{MCLR} 3V3
1. You can hide the "H1" labels on the mounting holes lol.

IshmanM:
Reply to @BluCodeGH:

Thanks for the new feedback. I've now addressed comments 1 through 8. 

I may have added an extra via once or twice to help avoid dead zones and accidental antennas, and add thermal relief.

Joe-Joe-Joe-Joe:
Not going to nitpick schematic at this point, but there are a couple housekeeping items 
1. Datasheet link for the AIS3624DQTR
2. Some of the caps and power symbols could be a bit neater, mostly the LDO and the floating cap on the accelerometer. Generally try to add a bit of wire between components, symbols, and netlabels so they are easier to distinguish, like this:
image

Onto actual stuff
1. Not _mandatory_, but I HIGHLY recommend test points. Power and ground this board maybe you can get away with probing your LDO or smth. But even if you just add a pad-type TP for one CAN line that will make your life much easier when you go to test
2. I think the routing and placement on the oscillator and C5/6 could be cleaner. Like, even if you don't do any big rearranging you could at the very least make the caps symmetric wrt the oscillator. See this figure from the datasheet
image
3. Dangling trace on F1 Pad 2
4. Not going to name them all, but best practice is to not have traces leave pads at an angle. C6 Pad 1 is a good example
5. Likewise, 90 deg trace junctions are also not good. e.g. MCLR on the back side
6. R4, C4, and C10 you have room to move further away from mounting holes. I didn't think about this either until last week but sometimes your mounting hardware gets tight.
7. This is again an FYI - we are talking about a board where you have a surface layer power pour. But the reason you put caps on voltage supplies is to bypass noise. In order to do that effectively, your cap needs to be physically as close to the supply pin as possible. C4 and C9 are doing literally nothing (and the trace from C9 to SD card pin 4 is redundant?). If you want them to do something, you would need to add a keepout area around the power pins and then route a trace through the cap pad to the pin

BluCodeGH (changes requested):
Looking good! In addition to Joe's comments I have a couple last things:

1. I'd recommend flipping your CAN connector around to the other side of the board. Generally you'll want to have one side be the 'front' in that you expect to have a bit more space/access there, so things like LEDs, the SD card, and the CAN connector should all go there. The programming connector is fine on the less accessible side since you access it from the edge anyways. Here you have the 'back' in kicad being the one you want more accessible, which is fine, but in the future its more clear if you make that the 'front' in kicad too.
1. The MCLR 3V3 text should be a big bigger and placed alongside the connector, like this:
image
1. R2 and R3 would probably be a bit comfier to solder if you spaced them just a little further apart.
1. It bugs me that C7 and C8 aren't aligned :)
1. I still think C3 will be easier to solder if you rotate it so its pins are facing in the same axis as U1's.
1. You'll also want to rotate C10 90 degrees, right now your soldering iron is going to also start melting the pin on U5 when you try to solder the upper pad.

IshmanM:
Reply to @Joe-Joe-Joe-Joe 

Thanks for the feedback, I've implemented it to the best extent possible without having to fully redo the layout.

Schematic stuff:
1. Done.
2. Done.

PCB stuff:
1. Added the one test point.
2. Symmetry added, but not fully rearranged because that would require moving many things.
3. Removed the trace.
4. Fixed.
5. Got rid of 90 degree junctions, often using Y shape junctions.
6. Created as much space as I can.
7. Now moved as physically close as possible. I don’t think it can be moved closer than this because the SD slot is wide.

----------------------------------------------------------------------------------------------------------------------------------

Reply to @BluCodeGH 

Thanks for the soldering points. I never considered those until you mentioned it.

1. Can’t do that since the accelerometer is on the KiCad front side too. Ill make sure to keep it in mind for future PCBs.
2. Done.
3. Done.
4. Aligned.
5. Done
6. Done

Joe-Joe-Joe-Joe (approved):
LGTM

DomenicHampson (changes requested):
Looks good address these comments and you have my approval.

1. Create a little more space between the crystal and the PIC. I'd also add a few gnd vias around the oscillator it's good practice and will help improve signal integrity. 

2. The pours don't go up to the edge of the board expand the zone so it does. 
image

Other than this LGTM thanks!

IshmanM:
Reply to @DomenicHampson 

Thanks for the feedback. I implemented it to the best extent possible given the limited space.

Joe-Joe-Joe-Joe (changes requested):
Payload requirements are being revisited. 99% we are going to want to break out some communication, analog/I2C/SPI, for a flowmeter and potentially for temp sensing. Board design shouldn't move forward until there is a formal meeting with payload and all this gets sorted out.

My understanding is that this is sort of the "custom payload board of the year" and the only other board in there (charging board) has enough interdependencies that we would not want to tack requirements onto it.

AshTheEngiqueer:
Looks pretty good, a few minor things:

1. 
image
avoids daisy chaining on the CANH hole of the Gecko Connector, which is what's currently happening here: 
image

2. 
I think the routing on the 3.3V lines for U3 is a bit messy? maybe something like this? 
image

3. 
Is there a reason you can't do something like this here? imo it's neater compared to what you have and is also shorter
image

4.
It might be easier/better to just do this as a trace rather than a trace to a filled zone?
image

5. 
This is also daisy chaining at the moment - also the traces are different widths? is that intended?
image

6.
This might be way too nitpicky but are we sure the spacing on these GND vias is right? 
image
From the work I've done on charging board, with a 12MHz oscillator, the via spacing should be 0.9mm.

7. 
I don't think you need two GND traces here? 
image

8.
The ground pour for the oscillator is a little bit scuffed because it's connected to the general ground on both sides - also having two oscillator ground pours is potentially scuffed? You can move the capacitors closer to the oscillator through-holes because the oscillator itself is on the front side, which then allows the ground plane to extend past the capacitors and makes it an undivided zone. Something like this: 
image
And then you'd want to separate the B.Cu Oscillator ground from the other GND plane on the back.

9.
This GND trace is doing nothing? 
image

10. 
This feels wrong to me for some reason - I think just running a trace to the adjacent via rather then having the filled zone and trace to the zone might be better, but I'm not sure on that one.
image

IshmanM:
Reply to @AshTheEngiqueer 

I addressed your comments. The only one I'm not sure about number 8, down below.

1. Daisy chain gone. Now using a Y shape. Just wondering in general why vias/holes need to avoid daisy chains?

2. Cleaned up.

3. Good idea. Done. I think I assumed there wasn’t enough space that when I was routing it.

4. It helps a bit with gnd plane continuity in that region of the board, so I’ll keep the zone. But I found a way to get rid of that odd thermal relief side trace I had before.

5. Daisy chain gone. The smaller trace goes to an op-amp, expecting lower current.

6. I think the current spacing is fine. Per the wavelength/20 rule of thumb I found, the vias are good as long as they’re less than 1.249 metres apart. Lol I think what I have now is also overkill, now that I did the calculation.
https://electronics.stackexchange.com/questions/359697/rf-via-fences-stitching-spacing

7. It’s just there for thermal relief and ground plane continuity/uniformity.

8. I made the zone undivided. I thought dual side gnd is fine. Is there a reason to remove/separate it?
oscgnd

9. Removed.

10. Maybe, so now I moved the via to the left to cover the end of the trace.

AshTheEngiqueer:
1) I'm not 100% sure if you need to avoid daisychaining on holes/vias - but you do have the space so my only thought was may as well? I could be wrong here though.
5) Gotcha.
6) Honestly that could very well be the case. I ran a quick calculation and assumed the answer was in mm because 0.9 meters didn't feel like it would have made sense to me - there's likely something I'm missing.
7) What I meant is "I think you can just do one ground trace to the via and exclude the filled zone from that area"
8) Dual side GND is fine, all I'm saying is that one side should be completely isolated from the main ground with a gap. See example from charging board: 
image
For this board, I *think* it's better for B.Cu to have the isolated GND plane because B.Cu has a few more other traces in close proximity.

IshmanM:
@AshTheEngiqueer 

5. Oh I see what you mean about the oscillator. Now I separated the B.cu gnd part.
4. I prefer to extend the filled zone for that area because it keeps the gnd more uniform between C1 an C3

Joe-Joe-Joe-Joe:
@AshTheEngiqueer @IshmanM 
for reference, you avoid daisy chaining pads so that if you damage/rip up a pad, you don't break the continuity of the trace as well. Though generally if you rip up a pad you have bigger issues

Ash and I fixed the oscillator ground plane scheme in most recent charging board commits - please see that PCB for the "ideal" way to route the oscillator plane. The planes should be isolated from the rest of the GND pour, but the pour on the same side as the PIC should be tied to GND only _through_ the GND pin on the PIC i.e. don't leave the oscillator GND plane floating

Joe-Joe-Joe-Joe:
Am I correct that the accel chip is QFN? if so
MAKE SURE YOU ORDER STENCIL WITH THE PCB

IshmanM:
@Joe-Joe-Joe-Joe 

Oscillator is now better. I'll mention to @JasonBrave that stencil needs to be ordered since he said he's gonna order the board

Joe-Joe-Joe-Joe:
1. Excellent work on the ground plane and caps. Only nit I have is the crystal is on the side opposite the caps/MCU. of course this is trivial to fix because you followed good design practice and didn't put any components on the other side :)
image
2. Please try to keep all silkscreen text in a uniform orientation
image
3. You forgot pullup resistors on your I2C lines. So have I. Super important.
https://learn.adafruit.com/working-with-i2c-devices/pull-up-resistors

Looks very good otherwise.

Joe-Joe-Joe-Joe:
Also, not directly related to the board but please figure out how you are going to harness the flow sensor.
Looks like it uses this specific connector, which requires specific Molex crimps (and probably a specific Molex crimping tool): https://www.digikey.ca/en/products/detail/molex/0510210600/242846?s=N4IgjCBcoLQBxVAYygMwIYBsDOBTANCAPZQDaIAnAAxUD8ArgC4C2A%2BtkfQE5K4C81KiAC6AX1GEATGRBUArGCqTFANhojRQA

I believe we should have parts for JSTs in the bay, but you or someone else should double check if we will have enough to support all our fabrication, and specifically if we have 6 pin since we don't have much that's that big.

JasonBrave (changes requested):
1. As joe said, add 4.7k pull-up resistors on both I2C lines
2. Add testpoints on I2C lines since JST are kinda hard to probe
3. There's no need to label the board as "V2"
4. The SD SPI signal name direction are from prospective of micro-controller right?

DomenicHampson (dismissed):
Hey Ishman looks great!

Address the above comments and you are good to go. Also don't forget to label the test points

Thanks!

IshmanM:
Thanks for the feedback. I addressed the comments:

@Joe-Joe-Joe-Joe 
1. oscillator flipped.
2. silkscreens rotated to be uniform.
3. i2c lines pulled up.

@JasonBrave 
1 & 2. The new i2c pullup resistors will serve as test points.
3. V2 changed to V1.
4. Yes, it's from microcontroller perspective. Signal name now adjusted to clarify.

@DomenicHampson 
- Test point labelled.

Joe-Joe-Joe-Joe (commented):
1. Something is up with your fills on the front side, DRC doesn't like it. It looks like that is a 3.3V zone beside a 3.3V zone so you can just delete it?
image

Otherwise LGTM

IshmanM:
@Joe-Joe-Joe-Joe 
I deleted it and extended the larger zone. DRC error should be fixed now

Joe-Joe-Joe-Joe (approved):

Joe-Joe-Joe-Joe:
Looks good. You can move to generating gerber files (let me know if you need help with that)

Joe-Joe-Joe-Joe (approved):
LGTM!

DomenicHampson (approved):

DomenicHampson (changes requested):
Actually before we order can you add test points for canl, the data lines for the accelerometer and the power rails. Also when you label the test points edit the existing TPx label. Do this for the canh point too.

Joe-Joe-Joe-Joe:
@DomenicHampson you should not edit the refdes labels directly (TPx etc). This is because if at any time in the future you need to "update from schematic" your changes will be overwritten, and you will have to relabel your TPs. I just put the refdes labels in the void outside the board edge which keeps kicad happy.

Also I didn't comment on the lack of a CANL test point because in 99% of cases if one signal is nominal both are, and in that 1% you can probe the IC. I suppose it doesn't matter either way considering the board has plenty of space.

JasonBrave (approved):
LGTM, I agree with joe's point

PR #76, opened 2024-01-19
added changes based on Zach's proposal

Tessa-P:
4 changes requested
-  add diodes to make batteries truly independent - they already have some
- make board resistant to failure of 5V rail - board functions fine already if 5V fails, we just don't get good current/voltage sensing
- Add RC sensing - done (using C of 4.7uF)
- add redundancy on pyro outputs - done

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/76)


DomenicHampson (changes requested):
Before I do a more thorough review…

Kicad is showing issues with the some of the old connector symbols. Can you drag in fresh symbols for the connectors
image
Also can you clean up the folder for RA. I’m seeing a lot of duplicates please delete them and the gerbers folder (since we I’ll be generating new ones for the board order). Also if I navigate to the altimeters sheet from the remote_arming schematic the changes you made aren't there. Please make sure the changes are also there too

Thanks!

DomenicHampson (changes requested):
Looking good thanks for getting those issues fixed! Just a couple small changes...

1. Can you add a note in the schematic showing how you determine the capacitor and resistor values for the filter.

2. Re-Annotate the schematic. Go through the electrical rules checker and make sure there are no duplicates

Thanks!

JasonBrave:
De-scoped for 2024

PR #77, opened 2024-03-24
Processor respin

Joe-Joe-Joe-Joe:
- Replaced reset button with DNP jumper for flight safety
- Added IMU connection (Power + USART1 + TIM1 CH1,2,3)
- Added JST to connect old ICM IMU externally and remove daughterboard mounting
- Put current sense amp on a valid ADC channel
- Added I2C pullups
- Made debugger connection 1.27mm to enable direct connect to STLINK-V3MINIE (please double check my pinout, I mirrored it realizing that I'm going male-> female but now I'm second guessing myself)

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/canhw/77)


Lucifersan:
I think the debugger connection pins are flipped, the other changes look fine

Joe-Joe-Joe-Joe:
I think it might need an external oscillator...

Joe-Joe-Joe-Joe:
The reason for the above change is primary to make UART and CAN timing more accurate. They both have demonstrated working, but we have had some intermittent issues with both and after a bit more research I've concluded that clock inaccuracy could be the source. I couldn't be bothered to/didn't have space to route a crystal and caps so you get an external oscillator

AshTheEngiqueer:
1. 
image
no-connection flags pls

2. 
According to JLCPCB this text will be illegible? They say min 40mil character height, this is about half that.
image

3.
Any reason this trace is off-centre from the capacitor pad? There's a few like this and I think you have the space to center a bunch of them.
image

4. 
There's a few scuffed silkscreens, like this one - I'm fairly sure you have the space to make sure they don't overlap with pads/mask
image

5. 
Daisychaining here and a few other places
image

6. missing datasheet on FB1

Looks pretty good to me overall though :)

Joe-Joe-Joe-Joe:
Especially after my experience at work, I am convinced that the IMU power input warrants TVS and reverse voltage protection. The 5V and ground voltage could be swapped at both ends of the processor-IMU harness, as well as at the processor board input connector (since the 5V line feeds the IMU output, rather than the 3.3V line. The drop over the Schottky diode is about ~250mV, (well within the input voltage range of the VN-200, and only 25mW of power dissipation @ max input current of ~100 mA)

Joe-Joe-Joe-Joe:
Also, I'm not going to add no connection flags to a 100 pin MCU. It makes sense on a connector where you need to communicate design intent, but I'm far too lazy to go one by one and add NC flags

ManavToor:
1. ERC is throwing an error here, why are these overlapping?
image
2. Vias on silkscreen (it's a tight board but if you can move them that would be nice)
image
3. Is this via large enough for the current passing through?
image
4. Since J4 is likely going to be facing upwards when we're working with the board, it might be better to move the LED's and test points to the back of the board as well
image

That's all that I could find. Looks pretty good

AshTheEngiqueer:
re: 1, DRC doesn't like the holes being so close together but that's just the footprint of the jumper. I had similar DRC "errors" with adjacent pads for a given footprint.

AshTheEngiqueer:
1. Still some daisychaining that I missed previously
image

2. Is there a reason this pad is attached to two vias? if it's because current reasons then the other trace off the resistor is probably sus? I think this is just daisychaining.
image

3. Just a bunch of daisychaining still
image
image
image


4. This trace is coming off at 90 degrees
image
as is this one
image


5. These silkscreens are a little bit sus (overlap with pad/via , you have the space to move them)
image
image
image
image
image
image
image

Joe-Joe-Joe-Joe:
Re: Manav
1. That was awful, but I eventually fixed it. Turns out setting a lower pad clearance than the board-wide minimum clearance just... doesn't warn you that you are being overridden by board settings
2. Just squeeeeeze em in there :)
image
3. Yes. That's the 5V supply to the IMU, not more than 200mA. The trace it branches from is fat because it has to carry external power (1A+) from the TP into the regulator circuit
4. Rearranged to put LEDs (and some silkscreen info) on the "back" side, which agree will probably end up being the outward facing side. TP courtyards don't fit on the bottom, but that's not an issue - we can wire in from other side just fine

Re: Ash
I think I caught em all, but I did change a few things around

JasonBrave (approved):

DomenicHampson (dismissed):
lgtm

BluCodeGH (dismissed):
Really well done, spotted a couple of small things:
1. The capacitor designators in the text notes by U1 and FB1 are wrong
2. Are you sure R4 and R5 are sized correctly?
3. I'm a bit nervous about the reset and restore jumpers being male (in terms of FOD), especially 1mm. Might be better to use a smd jumper footprint you can bridge with tweezers or a female header?
4. I didn't check your FET pinouts for Q1/Q3, pls make sure they are correct.
5. Daisy chaining C1
6. R6 is a bit close to the head of the bolt on the harwin connector
7. Is there a TVS diode in a more sane footprint? Also you probably want to put it right between the power and GND pins on the harwin right?
8. The footprint for the oscillator doesnt match the datasheet?

IshmanM:
I didn't go into the schematic, but here are some things that I noticed on the layout:

(The other comments brought up daisy chaining, so I won't mention it in detail. Daisy chaining the pullups on the SD card and other data lines may be risky, but things like daisy chaining to a decoupling cap or between sequential components in a power sequence are fine imo)

1. If you right-click and uncheck the visibility of those silkscreens that are outside the board, then I don't think they'll show up on the KiCad fab outputs. I imagine that's better in some way for manufacturing.
2. Acute angle on net to U4 NRST decoupling cap. There are also a few acute angles elsewhere. I don't think it's much of a problem for low speed or control signals but I'll point it out anyway.
3. Unsure, but the trace on the right below could cause manufacturing defects. Maybe route closer to the center of the pad, or at a 45 deg angle.
image
Another similar thing here:
image
4. GND discontinuity caught by DRC:
image
5. There are some thermal relief warnings. If needed, some of them are easy fixes, e.g. fix:
image
6. Some optional signal integrity stuff would be to minimize dead copper regions, e.g.: 
image
7. The power plane could have keep out regions at the board edges, to prevent stuff like this:
image
8. Ground plane has some spots where it just pokes and doesn't lead anywhere. Minimizing these or connecting the "tips" to the rest of the plane will lower noise pickup, if that's of any concern. E.g.:
image

I hope this helps

Joe-Joe-Joe-Joe:
Re: jack
1. Fixed
2. Eh you are right they are a bit chonky. I think I copied them from minisensor. I mean either way, they are known to work. I don't know what the nominal voltage drop or operating current for our particular LEDs is.
3. Those are not meant to be populated, but bridged as you described. I guess a surface pad would do the trick too. I think I just defaulted to this because that's how Kalman board did it (and my god did I have to use the kalman board reset jumper). Trivial to replace with pads if you think it makes sense
4. I double checked them
5. Fixed
6. Fixed
7. Does placement matter a lot? Anyway, I went with [this](https://www.digikey.ca/en/products/detail/kyocera-avx/GG020105100N2P/6826542), which is 0603. And I put it beside the datamate pins.
8. I linked the wrong datasheet. [Intended part is an ASE series](https://www.digikey.ca/en/products/detail/abracon-llc/ASE-4-000MHZ-L-C-T/1236872), for which the 3.2x2.5 mm footprint is correct.

Joe-Joe-Joe-Joe:
Re: ishman
0. Old me daisy chained the SD line pullups intentionally. Present me wants to think old me had a good reason for doing this other than vibes.
1. TIL. Done.
2. I changed whatever was practical
3. Those were ugly, I fixed them
4. Nice catch
5. I usually don't bother with those lol but I fixed the one you pointed out plus a couple other easy offenders
6. Done
7. Done
8. The ones i could fix by rearranging traces/vias (like the one you screenshotted), I did my best. But dropping 5 million keepouts everywhere is a pain.

AshTheEngiqueer (approved):
LGTM, good work :)

Lucifersan (approved):
Looks good :D

ManavToor (approved):
This bugs me, it's good practice to space apart your traces where you can
image

Other than that it all looks good

PR #81, opened 2024-05-11
Remote arming respin

ManavToor:
Guess who's back...remote arming!

Changes: 
- Removed extra shunt resistor on current sensing
- Split +BATT into +BATT1 and +BATT2, connected via a polyfuse
- Added current sensing for new +BATT2, connected to PIC RB5
- Killed all bypasses J5, JP1, JP2, JP8, JP9, JP10, JP11 
- Changed pull ups to connect to respective +BATTx instead of +5V with voltage divider
- Changed main and drogue connections to altimeter connections
- Moved stuff around slightly so schematic is more organized

Please check:
- shunt resistor values
- voltage divider resistor values
- i chose the right type of fuse and its value for batt connection
- pin outs for new altimeter connections
- i didn't delete smth i wasn't suppose to
- i didn't forget anything

please review so i can start the layout :)


zx100x100:
Looking good, I like the formatting!

A couple issues:

you are using +Batt in bunch of places, which means Batt1 and Batt2 are connected. nothing should use +Batt anymore, it should all be +Batt1 and +Batt2
image
image

Delete comment about jumpers that no longer exist: 
image

The resistance of your voltage divider here isn't quite right, not sure how you came up with these values? (also note previous comment about +Batt)
image
Let's do some math: Assume we want to stick with the maximum system voltage of 15.1V specified on the power sheet (due to the voltage dividers)
and the max voltage should be 5V at the divider: so 5/15.1 = R24/(R11+R24) = 0.33. Now a bit of a rule of thumb, but we don't want to have too low a resistance as it will result in excessive current draw, or too high as it will be very sensitive to noise or leakage current in the mosfet. It's good to use common values, so lets try out R11 = 5k, R24= 2k, this gives us:
R24/(R11+R24) = 0.286
@15.1V -> 4.314V (max)
@9V -> 2.5V (nominal)
@6V -> 1.7V (minimum (a  bit arbitrary)
Checking the MOSFET datasheet: https://www.diodes.com/assets/Datasheets/DMG2302UK.pdf
Minimum Vgs to turn on the MOSFET is 1V, so even at a very low battery voltage of 6V, this should work.

Since this is an important system lets check a few more things:
Maximum Current draw for the PIC to pull the pin low: I = V/R = 15.1/(5k + 270) = 2.87 mA (you can check the PIC datasheet but this is very reasonable)

Effect of leakage current on divider voltage:
Igs max = 10uA from datasheet:
V = IR = 10uA*5k = 50mV (this is negligible)

Max Continuous current drain from battery:
I = V/R = 15.1V/(2k+5k) = 2mA. this should be reasonable given this draw only occurs when the batteries are turned on (batteries are turned on just before tower raising, probably no more than 5h before launch)

So it looks like R11 = 5k, R24 = 2K should work well, but make sure the above makes sense and I didn't make any mistakes.

Other than that I think it looks good!

ManavToor:
RE:
1. Yeah, libraries were very glitchy so BATT1 and BATT2 both apeared as BATT, should be fixed now
2. Footprint for the fuse shouldn't change so we can update the value later
3. I had assumed 9V not 15V for the divider, went over your math, makes sense and checks out

UPDATE:
1. Began the routing, kept the large components in the same area, just moved some connectors over
2. Added GND copper pours to top and bottom layer
3. Everything else has been moved around
4. Changed 30mil traces to 40mil, kept all other trace sizing the same

Once fuse values are finalized I'll update the schematic pdf

zx100x100:
Looks good, a few small things:

Apologies I missed this in the first round, but minor thing: lets have the fuse between raw_bat1 and raw_bat2. this will allow us to have current sensing for each altimeter, which is a bit more useful than for each battery.
image

these traces currently daisy chain through one pad, avoid that 
image

this doesn't follow best practices for routing, but looks hard to fix so it's probably fine 
image

I think this is inherited from the old desing, but these are all high current traces, can you thicken them up as much as possible? 
image

the routing of the pull up jumpers is particularly concerning, since it is daisy chained through the buzzer contacts. these are critical, high current traces, please re-route: 
image

far less critical, but more daisy-chaining 
image
image
image
image
image
image
image
image
image

ManavToor:
Removed the daisy chaining

Moved fuse to raw_bats

Beefed up pyro traces

Artem-Sotnikov:
Overall - good. From the electrical side, I'd say it's ready for ordering.

Only nitpick is that the silk screen placement here is questionable and will be hard to read. There is plenty of space to position the silkscreens in a less cramped location near these components.

image

zx100x100 (changes requested):
Pull up traces are still very thin, this is a critical issue and needs to be fixed: 
image

Tiny nitpic: this comment still needs to be deleted, it doesn't make sense cause the jumper was deleted

I'll try to do a more detailed review later

Joe-Joe-Joe-Joe (approved):
I have no idea how the altimeters work, or how these jumpers are designed to work. At the very least, naming J5 "A1 Connector" and J8 "A1 Connections" is just super confusing.

Also your Q4/5 labels are rotated
image

Everything looks good from an electrical standpoint except the jumper trace width that Zach already identified

ManavToor:
not sure how i missed those.

what comment are you talking about? The one regarding the pull-ups?

Fixed labels

zx100x100 (approved):
Looks good! 
Only important comment: did we confirm that this has the newer mag switch footprints? have we ordered more mag switches?

Gerbers look good

Minor note for future things: when you have these sketchy fill zones that barely connect the part I like to manually route a trace too:
image

you have a few slightly messy silkscreen overlaps, but not worth fixing.

For future re-spins: the 5V power plane is a bit excessive, most of it doesn't connect anywhere, so it could probably be trimmed and much of it replaced with another ground plane, but this is also not worth changing.

PR #83, opened 2024-06-20
voltage divider changed

Lucifersan:
I just changed the resistor dividers


AshTheEngiqueer (changes requested):
Where are changes? assuming you're talking about the divider on the +BATT voltage sensing line?

Lucifersan:
yes, because with the old divider, if the 24V line shorted to the batt line it would send like 12V to the pic, which is not great. Now if 24V line shorts to bat or VCC it doesnt matter and the pic gets 5V input anyways. we lose a some fidelity this way but at the rate we've been killing pics on this board, I think that is an acceptable sacrifice to make.

AshTheEngiqueer (approved):

Joe-Joe-Joe-Joe (approved):
LGTM

Joe-Joe-Joe-Joe (changes requested):
oops
just realized you did not regen the schematic PDF
that would be bad

ManavToor (approved):

Lucifersan:
pdf regenerated

Joe-Joe-Joe-Joe (approved):
lgtm

Pdada1 (approved):

BluCodeGH (approved):

PR #84, opened 2024-06-27
Usb debug respin respin 

miles-milne:
Do a thorough review this time pls and ty


icanc0:
have you settled on the board thickness yet? you'd also want gnd around the differential pair on the same layer, and make that as close to the differential pair as you can. Use the JLC calculator with the smallest trace to trace and trace to copper values. 

https://jlcpcb.com/pcb-impedance-calculator

image

PR #88, opened 2024-09-06
2024 cleanup

JasonBrave:


ManavToor (changes requested):
keep logger

ManavToor (approved):

Rio6:
why are we removing harwin footprints and 3d models?

JasonBrave:
Because they got migrated to here because there are other repo need some of the footprints https://github.com/waterloo-rocketry/kicad-footprints

Rio6 (approved):
LGTM we can always get it back if we need something

miles-milne (approved):
Looks good to me. Happy that symbols/footprints are being moved to a git submodule.

PR #90, opened 2024-11-02
Processor 2025 Full Review

Armaan-Sengupta:


Joe-Joe-Joe-Joe:
1. Why is this called "spi_addr"? Shouldn't it be "I2C_addr" or smth?
image
2. For net labels beside hierarchical ones they are not technically overlapping but still very hard to read. Please extend one as has been done for CAN Tx/Rx above
image
3. This is my failure from before but this should live in the whitespace on the right side and have a net label to PH0
image
4. Please point power symbols up and grounds down always
image
5. What....
image
6. Is this fuse value on 5V still accurate? Same with R1. Have you actually done your power draw calcs somewhere? If you need a reference,[ "processor board current"](https://docs.google.com/spreadsheets/d/1yrRCH4vebvrrPBrgjtCP7KK3ymbRddPD5F0F_Y44o-U/edit?gid=1765370054#gid=1765370054)
image
7. A detailed review of the buck and the LSM will follow, I haven't looked at these in detail yet

Joe-Joe-Joe-Joe:
7. When you do get to layout, make sure you annotate the axis directions of the LSM clearly on the silkscreen. Also note that their "X" and "roll" axes are different, which is not the same as our (well, my, but nonetheless the correct one) axis convention for the rocket
8. What does Movella Sync In/Sync Out do? Does it make sense to connect it to timer pins (I think you have those free somewhere) 
9. D:
image
10. Inductor packages tend to be nonstandard. Have you selected a specific inductor for L1? There is a datasheet section on inductor selection which you should read if you haven't already.

AshTheEngiqueer:
1. Pin 6 is a NC pin for the 2568? If so, no point in including it in the symbol.
2. Max current on the buck is 1A - are you sure this is sufficient? Have you double checked your current draw maximums? It seems pretty close from my review of the controls 2024 spreadsheet and the changes made, but I'd double check.
3. nit but the datasheet on U5 links to the page on stm, rather than the datasheet pdf
4. Not super familliar with I2C but what's the justification for the DNP resistors R16/17/19/21?
5. also nit but there's some inconsistent formatting with resistor values - there are 10k resistors marked as 10kΩ or 10kR. afaik team standard is 10k, same with capacitors (1u vs 1uF)
6. Bold to assume you can find an inductor with a standard footprint - take a look on digikey for the right value and see what you find, pick something and use it's footprint. We ran into issues with finding standard footprint inductors with charging board last year, and I suspect you'll find the same this year.

Armaan-Sengupta:
@Joe-Joe-Joe-Joe :
1. Yes sry, too much SPI at work
2. ty
3. yea good idea, didnt notice that was the only connection lol
4. 👍
5. test that you are really getting 3.3v before it powers the entire system up potentially breaking everything if it is not 3.3v (ex. wrong resistors for voltage divider), then just put a blob of solder there to connect
6. I have not, just carried over, thanks for catching that. But [upon checking](https://docs.google.com/spreadsheets/d/1HMfB6qHngxaIYGLUWItAfbt1gtbMd2xDmdOFUf7_RjQ/edit?usp=sharing) it should be fine, buck allows current to be a bit lower, but that should be fine.
7. Good idea, I will follow the sensor's coordinate frame rather than our team standard
8. I dont think they will be required, they are there primarily because harwin is 6 pin, lol. I think we are most likely to use the Sync In if we want to specifcally request data at that instant. Yea I can tie sync in to a clock line, I still think we will probably use it as a regular GPIO but doesnt hurt. 
9. Fixed
10. I did, but I would like to read that resource to check against, can you send it?

@CasualIntellectual :

1. Its a vendor part, id rather just leave it to match the datasheet, and so we dont need to modify the provided file
2. Yea, it will be less than 1A, calcs linked above
3. Fixed
4. We can place them if we want, resistors in parallel lower the equivalent resistance (weaker pullup), RC time constant goes down, we can run I2C faster if we want
5. Fair enoug for resistors, just thought it looked nicer, but removed👍 . For caps it looks really strange IMO, but if its fr a team standard we follow I can remove it (same for inductor ig).
6. I did check, and just oversized, I think its too oversized, ill learn how to make a custom footprint during routing

Armaan-Sengupta:
> 10\. datasheet section on inductor selection

Oh I assumed this would not live in a specific Inductor's datasheet, but rather be some kind of general third party guide

AshTheEngiqueer:
1. This via is only connected on one side, I think you can just remove it? 
image

2. aaaaaahhhhhh what is this daisychaining
image

3. Is there a reason why all of your test points are through-hole?

4. There's a lot of pad-silkscreen overlap in general
5. Test points not labeled?

Armaan-Sengupta:
1. Good catch
2. Good point ty
3. Not all, the I2C ones are not, where I had space to fit them, I made them through hole since its more convenient
4. Yup hadnt gotten to cleaning that up yet, you got the jump on me before I PR'd this lol (but ty I appreciate the early feedback)
5. Which ones? Most are looks like I missed the 5v and gnd inputs, still need to work on silkscreen stuff

Armaan-Sengupta:
@Joe-Joe-Joe-Joe @AshTheEngiqueer @JasonBrave full PCB ready for review

Joe-Joe-Joe-Joe:
1. Round your corners
image
2. Why are the hole spacings nonstandard? Should be in increments of 0.5in
3. I don't have time to dig through it all but you should think about how your harness connections are oriented. Right now, if someone connects the rocketCAN harness to the pololu connection inadvertently, they will expose the 3v3 line to -5V and that's generally bad
image
4. Nit, but I would prefer "RocketCAN" over "Main Harness". Less ambiguous
5. What is this?
image
6. Avoid daisy-chaining pads (yes this might have been me lol)
image
7. SDMMC_D0 should be length matched to the other data and clock lines. Something like this is easy enough
image

Armaan-Sengupta:
1. yea was planning to do at very end in case any modifications required changing board outline (does anyone know how not to make this a PITA once fillets are added?). If no one else has anything in a day ill consider it final and fillet the corners
2. oh they are just in mm, inches moment, sigh okay I forgot we do imperial here lol
3. Yea I like that idea, ty. Btw In specific, the old setup that scenario would have resulted in 5V being connected to gnd (not 3.3v):
image
4. Yea I thought since it was also main power supply, but nvm I like that more asw
5. Good catch ty
6. Yup good catch missed it ty
7. Oh damn speedy boy, 24Mhz, didnt realize, will do.

AshTheEngiqueer:
1. Daisychaining
image
2. 
image
3. Still no rounded corners?
image
4. 90deg trace aaaa
image

In general looks alright apart from lots of daisychaining/bad practice (traces out of pads at 45deg angles, or way off center, or just interesting routing decisions)

Redstone-ray:
Is Daisychaining bad practise? I don't understand.

Rio6:
If you daisy chain and someone skill issued during soldering and ruined the pad, the entire trace is ruined.

There are arguments for and again the 45deg trace but in the end is purely for aesthetic for us like rounded corners.

Armaan-Sengupta:
I was told by someone way back in my first coop when I first designed a PCB that 90 degree traces are bad for boards with extremely fast signal speeds, and have generally just stuck with that, but in reality I think it does not rrly matter, ig I just missed this one as the 3 way joint let it elude me, will fix.

Yea I keep getting caught on the daisy chaining stuff, lol ty, I do generally try not to daisy chain, because yea like Rio said if someone ripped the pad of (which I think is actually not too uncommon), the entire "bus" is ruined. I will make sure to go through the entire board not just the stuff I worked on to check for it 👍.

Armaan-Sengupta:
Dang it okay, ill modify the other two instead. We plan to JLC assembly for as much as is possible, I am currently trying to figure out that side of things. Yea I think so, Jason mentioned if its processor specific it stays with processor. Yea idk what files are needed vs are not, I think most of those were autogenerated?

Armaan-Sengupta:
Harwin connections flipped to other side of bord as [per Jason request](https://waterloorocketry.slack.com/archives/C05EUQLFP4J/p1731895567948139?thread_ts=1731866155.828359&cid=C05EUQLFP4J), and pin connections re-done as per [Joe's comment.](https://github.com/waterloo-rocketry/canhw/pull/90#issuecomment-2481429888) change log of harness connections [here](https://docs.google.com/document/d/1ESQr2ki7eqf_K064cuU7I857oRs_P2dz_hfp1vAU-G0/edit?usp=sharing).

Some cosmetic changes now that we know the board will be mounted flipped. 

Final step before production is for me to source matching components on JLCPCB and setup all the PCBA files correctly (BOM + CPL) but in terms of the PCB itself, it is done, please if you have reviewed it, mark it as approved so we can merge, else kindly review it first. Thanks for the help yall.

Redstone-ray (changes requested):
1. What is this, where is this going
image
2. b b b b b b b b b
image
3. Where is this x pin going
image
4. woah 90 degrees
image

Joe-Joe-Joe-Joe:
So, I still have some reservations about this. I understand flipping the rocketCAN connector to mate with the new sled. But if you are flipping the other connectors does that mean they will also mate with the sled, and you will plug the external IMUs into the sled as well? Are you going to try to connect them directly and you just need to meet the clearance requirement? I am very lost as to how this is coming together.

Joe-Joe-Joe-Joe:
Also fwiw I won't approve it until you push the final production files

Joe-Joe-Joe-Joe:
@Redstone-ray from the electrical standard:
image
In this case they are all GPIOs and it doesn't matter either way but tying NC pins to ground is absolutely not good practice

Armaan-Sengupta:
> @Redstone-ray from the electrical standard:

For future reference. Joe's comment is regarding the standard practise for NC pins is for motor controller board re: [richards comment](https://github.com/waterloo-rocketry/controls-2025-electrical/pull/1#pullrequestreview-2444295234). It is not related to processor.

Armaan-Sengupta:
> So, I still have some reservations about this. I understand flipping the rocketCAN connector to mate with the new sled. But if you are flipping the other connectors does that mean they will also mate with the sled, and you will plug the external IMUs into the sled as well? Are you going to try to connect them directly and you just need to meet the clearance requirement? I am very lost as to how this is coming together.

@JasonBrave idk about how recelec looks, you told me constraints for mounting I simply followed, if you could answer Joes questions that would be great

Redstone-ray (dismissed):

Redstone-ray:
Please confirm manufacturing details and ST IMU can be assembled before ordering

rexblade21:
image
Why is this routed out and then connected to the 5V pour with vias as opposed to connecting it directly to the 5V pour on the top layer?

rexblade21 (approved):

Redstone-ray (approved):
Idk why you just copper pour this or run a trace out instead of via it from power but sure can't give two less cents of fucks
image

Joe-Joe-Joe-Joe:
You need to regenerate the drill files (make sure they replace the ones in `gerbers/`, the default puts them in `production/`

Joe-Joe-Joe-Joe:
JLC has a specific format for gerbers/drill files, make sure you follow it

https://jlcpcb.com/help/article/how-to-generate-gerber-and-drill-files-in-kicad-8

Redstone-ray (approved):

Armaan-Sengupta:
Deleted those gerber files they were stale and from the old way of manually doing it, all relevant production files are in production folder, and all gerbers are in the zip file which you can directly upload to JLCPCB.

AshTheEngiqueer (approved):
LGTM, good job 👍

rexblade21 (approved):

PR #91, opened 2024-11-19
SRAD Altimeter

StarlightDescender:
Currently a schematic review only.


KavinSatheeskumar:
First round of comments, going to get food and will keep roasting

- Termination Resistor for GPS CAN Transceiver
- Talk to the GPS via Serial (or something else) rather than CAN, since having two independent CAN buses on one board has been a known pain point for Payload 2022
- can't look at the web-bench TI link because I cannot login
- Add no-connect to unconnected pin on the STM
- Looking at the TLP3543A here, (according to Ash), the direction of current of the two mosfets must be opposite, however there is nothing on the schematic to distinguish that. (IE, the little diodes inside the symbol both point in the same direction and the part numbers are the same)
image
- please add data sheet to all IC's used in your schematic

KavinSatheeskumar:
More comments

- This should be 3V3
image
- Where do GPSCANH and GPSCANL
- This dosen't match the label coming into the sheet (from the outside its PYRO, on the inside its +PYRO)
image
- Connect to Ground 
image

Armaan-Sengupta:
1. Add another filter capacitor to 3v3 according to datasheet (same for the GPS one)
image 
image
2. VDD pin should be connected to 5V for GPS CAN
image
3. Enable should not be floating tie it to ground
image

Unfortunately i'll need to be given some context about how the pyros work before I can review those parts, but otherwise lgtm.

Armaan-Sengupta (changes requested):
https://github.com/waterloo-rocketry/canhw/pull/91#issuecomment-2579103923

AshTheEngiqueer:
@Armaan-Sengupta 
1. sorry, wdym by another filter capacitor according to the datasheet? the image you've suggested says I need one on VDD and VIO?
2. I'm not convinced it should be, the GPS module takes 3V3 in so that's my justification behind running the altimeter/gps CAN on 3V3
3. datasheet says it doesnt matter

Armaan-Sengupta:
1. Yea to add a 0.1uF cap to VIO (3v3)
2. So the VIO should be 3V3 then right? But the actual MCP needs 5v (page 9) so VDD should be 5V
image
3. Ah yes I read the min max voltage backwards but I see the note on page 3 now. Also it has an internal pullup which I did not see before, yea chilling what u did makes sense.

Rio6:
1. D1 is 1A max, is that enough? 圖片
2. Is this huge capacitor neccessary? With LiPo instead of 9V I'd imagine it can source a lot more current on 12V side, and keeping 3V happy. 
圖片
4. These fuses can probably be bigger, what is the limiting factor here? 
圖片
5. 500ohm gives (3.3V - 1.8V) / 500 = 3mA which is just above the trigger current. The recommended forward current is 10mA from datasheet. 
圖片
6. Fix DRC (both errors and warnings), even if most of them are harmless they can bury actual errors

AshTheEngiqueer:
1. This was the diode sourced by Jason so idk, I'll look
2. Yup, good point, will think about this one more
3. Limiting factor is we'd like the fuse to break if the ematches are shorted (but this is kind of fake now that we have the 4R resistors I think? mag switch can take 20A for 5 seconds and 4R limits us to ~3A in case ematches are shorted)
4. Good catch, I forgot to account for the SSR forward voltage drop.
5. Not sure how to fix a bunch of the warnings? stuff like "input power pin not driven by output power pin", or "symbol pin or wire end off connection grid". those I've ignored, added NC symbols to all unused pins.

Rio6:
For "wire end off grid" you can right click and ignore all that type of error.

For "power pin not driven" there's a power flag component you can connect to the power nets like this
圖片

zx100x100:
1. pedantic, but I would recommend putting the current sensing math on the schematic so it is obvious the max current for each channel. I also don't love separating the shunt from the sensing amp on the schematic since it might be easier to forget that they need to be routed very close together.
image

2. What is your math for these pull-ups and sense circuits? what do you expect to see on the sense line across the different operating conditions? (IDK if it does what you want it to do)
image

3. what is this supposed to do? it looks like current sensing, but the E-matches have a non-linear, variable resistance, so this doesn't work for current sensing since you don't know R
image

4. this part number is a 6 pin connector but the schematic is 2 pins! based on the spreadsheet 6 pins is correct, in which case I would recommend redundant connections for + and - so that we can harness parallel E-matches separately if desired.
image

5.As discussed documentation really needs updates, particularly to requirements, and description of system architecture

BluCodeGH (commented):
1. Your labels for the programming pins are wrong. When only 2 pins are connecting we're using SWD not JTAG mode
2. Why do you have boot0 selection and a reset jumper? IIRC we've never used either, and camera worked just fine without either.
3. Why do you have an external oscillator? Again camera worked without it, the internal one is usually fine?
4. Resize the LED resistors for 3.3V
5. There are weird doubled up DNP resistors on the pololu sheet?
6. You don't need to connect to the pololu addr pin, save yourself the hassle and 1x1 connector.
7. Also you should inject 3.3V direct to VDD on the pololu not via VIN
8. It would be nice to use the 4R resistor as a pyro current sense measurement for real, probably via a voltage divider of the low side into the STM?
9. CANH/CANL test points are kinda silly cause there's a connector right there
10. I'd move the main/drogue testpoints to the pyro sheet
11. I'd also move the current sense testpoints to the power sheet
12. The INA4180 is kinda hard to lay out, I'd suggest 2x INA2180, one on the power sheet and one on the main sheet
13. The 5V LDO max input voltage is 12V, but lipos go to 12.6
14. The 3.3V buck has a soft start time of 2ms which isnt enough to charge the capacitor given its 3A limit, so we're going to be overcurrenting it. I really think you should reduce the requirements on that capacitor to make it a more reasonable size, eg maybe you only need 20ms and the STM32 can handle going to 2V or something, idk. Otherwise you need a buck with configurable soft start.
15. F1 has the wrong footprint, also are there smaller options? Its kinda big
16. L1 schematic value should be the part number, not the size
17. Same for C31
18. The ferrite bead is 0603, choose an 0805 one?
19. The SDMMC app note link is missing
20. I glanced at it but pls try to configure all your pins in STM32cube and make sure its happy before ordering
21. Like Zach mentioned you should change the pyro connectors, I think you should have a single M80 harwin with main + drogue in the same pinouy as RA.

AshTheEngiqueer:
@BluCodeGH 
2. Processor had both, I'll ask Joe if either was useful.
3. I know we use external oscillators for the PICs and processor also used an external oscillator. Didn't realize camera was fine without.
15. Not really? if we want a PTC with a hold current of 7.5A, max voltage >= 12.6V, and SMD then this is our only option. I suppose THT might be better, I'll do some looking

BluCodeGH (approved):
Schematic LGTM pending a few tweaks:
1.  Pyro connector pinout is wrong, we want main on pins 1/4 and drogue on pins 3/6
1. Typically you should put the current sense shunt resistors with the main circuit rather than near the INA, makes it clearer what's going on
1. Why do you have the pair of FETs that switch MAG_VOUT to +PYRO? The point of the pyro circuit is that it requires two actions to fire so there doesn't need to be any "armed" state. (Also those big FETs are really annoying to solder)
1. You should set the REF of the pyro op amp to 3.3/2 volts so you can sense current flowing in the opposite direction too right? (eg PY+ shorted to GND). You might need an analog input on the REF voltage too to be able to accurately determine the zero value, I'm not sure?
1. The gain of the pyro op amp might be too high? Assuming a single 1R pyro I get 6 mA through it (when we test by turning on the high side) which is 6V output at 1000V/V?

I didn't go through the full failure mode state analysis, do make sure it's up-to-date given the pyro current sensing you added and whatever opamp changes you make.

AshTheEngiqueer:
1. fixed
2. Zach and you have opposite opinions on this one - his argument is that the current sense resistor needs to be placed close to the INA
3. Yup, fair point. I was thinking about how RA worked and was wanting to build that functionality into this altimeter, but I see your point about how that doesnt make sense.
4. Do we care about being able to see if current is flowing backward with that op amp? If PY+ is shorted to some lower potential point (GND, for example), then SENSE_V will drop very close to GND as well.
5. Good catch

Pdada1:
image
make this label more descriptive. Was looking at the gnss breakout board and it's hard to tell what this means

Joe-Joe-Joe-Joe:
1. We have killed processor boards plugging rocketcan into the wrong gecko mistakenly. I wonder if there is a way to make this more idiot-proof
image
2. GPS datasheet? No way of knowing if this pinout is right
image
3. doesn't match Webench?
image

StarlightDescender:
1. Unfortunately no way to make this more idiotproof, if you plug powered rocketCAN into the GPS connector you're sending 12V somewhere.
2. Also this pinout is arbitrary, just has to match the breakout board and connect to the right pins of the 20-pin header.

StarlightDescender:
also oops thats embarrassing, the text link was correct but it hyperlinked to another converter

Joe-Joe-Joe-Joe:
1. Current limiting resistors on SSRs are skirting the minimum forward current requirement pretty close (about 6 mA vs 5 mA). For nominal forward voltage and current (1.64V 10 mA) you want ~166R. I dont know off the top of my head how much current the STM can sink/source. Is that related?
image
2. I question whether the power saving is worth using yet another different amp
image
3. I feel like this should be 0.1uF. Same with most other decoupling caps. We typically use 10uF only for decoupling the 2562s, I think because the CAN lines get switched so much. I'm no expert in why smaller caps are used, but I think it has smth to do with internal inductance?
image
4. Nit, but please exclude holes and (pad) TPs from the BOM e.g.
image


Layout
1. uhhhhh
image
2. 
image
image
3. Please clearly mark the clocking of the programmer on the silkscreen and ensure you have clearance to plug it it, given that clocking
image
4. Would be best to model the polulu in board CAD since it is a significant protrusion. Same w/ magswitch and battery connection
image
5. We talked about how this is designed to produce a unique voltage for each failure mode. Are those in a table somewhere? dont want telem doing MTE120 practice at the ops table figuring out whats failed
image
6. The heat sink plane for the power resistors is isolated. You cannot dissipate 35W continuously like that. What is the longest sustained power dissipation case here?
image
7. Reminder to specify your part numbers if your parts come in different packages. I almost ordered the SSRs in a through-hole package
image
8. Zone border could use some cleanup, there is this random strip just on In1. Id just extend them all to board edge
image
9. Nit but name these GND/PWR
image
10. TBH this Ihsan guy seems to know way more than I do about laying out power supplies I'm just gonna wait for him on that
11. Feel like this has clearance to clean up the layout a bit
image
12. TP 2 and 4 should have descriptions. For TPs I recommend moving the refdes off the board edge and then adding static text to describe it.
image

ManavToor (changes requested):
Reviewing this on my lunch break at work lol, looks very good. Few small notes:
- you have a good amount of DRC errors and I think your mounting holes might be too close to the edge
- Inductor for buck converter is pretty close to a lot of the current sense stuff 
image
- Your grounds are not connected, this entire buck converter circuit is floating
image

ManavToor (approved):

ihsan-sa (approved):
LGTM

One thing I noticed: DRC is flagging a few incomplete thermal relief connections between planes and components. Should be fine because there's prob very little current, but just an FYI. In some places, you added an additional trace and via to the net.

Also, silkscreen is flagged for being too small in some places as per DRC, but that's not critical either.

Otherwise looks good! :)

Lucifersan (approved):
LGTM as well! No data/clock lines swapped and all that :heart_eyes:

PR #92, opened 2024-11-20
Injector sensor hub

ecslewis:


Pdada1 (changes requested):
image
With a 63 mR you only get span to 2.2V when you can go to 3.3 V on the input. You can use a 91mR instead to get more span to 3.18V and use more of the measurement range. 
image
Label the frequency of your oscillator
image
Don't use global labels, use net labels instead here.

Hexamech:
1. Invalid datasheet link for buck converter
2. Minor but rename MC2562 to be MC**P**2562
3. Can connector and Programming header footprints seem to be in an uknown library for me

Pdada1 (commented):
image
I think your still making changes so I'm gonna wait go more thouroughly through the schematic, but something I noticed is that your connector is backwards. You wanna connect the wires to the round part (red arrow)

Might also be nice to list the sensors the board is being designed around in the analog sensor sheet. 
image
Also your missing a voltage connection for this LED

Pdada1 (changes requested):
K, like was me and ash were talking about, this board isn't actually running any sensor on 3V3 and every other component can run on 5V supplied from the can bus. So you can get rid of the LDO (SPX 3819) and shift everything to 5V. That'll mean you have resize the current sense resistors, as well as the current limiting resistors on the LEDs, as well as replace other connections
image
voltage symbols should in general point up
Might be nice to list the sensors the board was designed for somewhere on the schematic

StarlightDescender (dismissed):
^ see above, and also please revert your changes to propulsion board + daq can support

Pdada1 (dismissed):
First off, I apologize for taking so long. 
image
image
Your Can connector gecko is reversed. Remember to check the number each pin is wired to and not just the position since the connector symbol is the same when mirrored but each number corresponds to a different physical pin on the gecko. Second image is connector from rocket_power with right pin number connections
image
Now that the pic is running on 5V it's max adc voltage is also 5V so you should change the current sense resistor to take advantage of this additional range. at you can cahnge the 62mr to 130mR and get span to 4.5V instead of roughly 2V
image
Nit but you can get rid fo this extra bend
image
Nit but you should connect these NC flags directly. The reason your having trouble placing them ontop of the pin is because your grid size has changed at some point. 
image
If you change it to one size down by right click menu, you'll be able to place them directly
Otherwise LGTM
@StarlightDescender @JasonBrave 
do we have the can harness assignment documented somewhere, or maybe we can throw it in electrical standards. It seems really jank to have to open existing boards in order to check against

StarlightDescender:
We do have CAN harness pinout, but it's out of date 💀 to be updated Soon (tm)

Pdada1:
image
please point ground downwards otherwise LGTM

Joe-Joe-Joe-Joe:
1. You should only have one KiCAD project file. You seem to have 3 plus another nested project. Delete the excess ones please
image
2. Your hierarchical labels are not consistent. You have 6 on the MCU, 8 going into the MCU sheet, and only 6 coming out of the sensors sheet (which themselves are duplicates)
image
image
image
3. We normally stock red, green, and blue (and white) LEDs. Is there a particular reason this needs a yellow one?
image

ManavToor:
Please address Joe's comments and also begin working on the layout

StarlightDescender (approved):

Lucifersan:
Given the nit that ash did I would also say another nit is while you're doing that, please move the LEDs (and their respective resistors) a bit further from each other, they're really close right now. We're not tight on space so there's no reason to cram them like that imo

Lucifersan:
Screenshot 2025-05-27 at 7 36 01 PM
In a similar vein, another nit: I think these connectors are both 1) _really_ close together for no reason because theres space to move them apart and 2) not evenly spaced which is a bit :(

Lucifersan (approved):
I will not block on any of these nits because of the time pressure, but I do think they should be changed regardless

Lucifersan:
^ seconded lol

Lucifersan (approved):
lgtm :D

StarlightDescender (approved):
send it

Pdada1 (commented):
image
nit we have daisy chaining here
image
 nit you can make this trace smaller by assigning this to pin 15 instead
image
trace can be eliminated
image
think it's kinda sus to have the 5V islands be bridged by the pic 5V pin, and also have that super in area on the pour
image
there is a lot of bplane area here to put in some stictching

Pdada1:
> nit but board name should be "sensor? I hardly know 'er!" rather than charger. otherwise LGTM

I honestly don't know how much the first years care, but I think they should be able to choose the name for the board. How else are they supposed to establish a series

Pdada1:
image
We want the both planes of the oscillator to be isolated right?

Joe-Joe-Joe-Joe (approved):
LGTM

PR #93, opened 2024-11-26
Rocket power board

Pdada1:


AshTheEngiqueer:
Assuming this is good to review despite the PR still being marked as a draft?
1. There's a lot of stuff on the root page, and the subsheets are basically empty. I think you could have your buck, battery charge circuitry, and current sensing on the same sheet? In general there's some messy stuff that would benefit from being tidied up, both for aesthetics but also readability.
2. You should note the fuse current, and on that note I think 6.3A is too high? my napkin calculation says that 4A is sufficient - 2A max through the gecko on both the 5V and 12V lines, and the battery current draw due to the 5V is at most (2A * (5V/12V) * 1/e) = 0.92A (assuming an efficiency of 90%).
3. On the 5V enable, I'm not sure why you're using the Drain-Source voltage to size R14/15? Should you not be using the Gate-Source Threshold voltage to calculate the divider such that V(HIGH) across the divider gives a voltage just above VGSth?
4. Your P-MOSFET has an insufficient Continuous Body Diode Forward Current (1.5A) - from the buck and the gecko we can output up to 2A, is there a reason we're limiting it here?
5. Is there a reason you don't have a low-pass RC filter on the INA2180? You should also make a note of the gain of the INA.
6. Rather than add a solder jumper to the CAN termination resistor, you could mark it as DNP and only add it if you do actually need it.

ManavToor:
Please organize your schematic better, it is very difficult to read.

AshTheEngiqueer (dismissed):
1. Sorry I wasn't clear, I meant that you should put a differential low pass filter on the *inputs* to the INA.
nvm im trolling, yes the low pass filter should be right at the ADC pin
2. How did you size the fuse? What math/current draw/data has been referenced?
3. Current sense resistors not sized still?
4. Some of your footprint links are broken - There's a footprint from tutorial board, as well as several from other projects and some from folders in your downloads folder. These should all be in the project specific footprint library, and should be referenced with a relative filepath, rather than an absolute filepath.

AshTheEngiqueer:
Still some weird schematic stuff - ex. lipo voltage from power subsheet goes to VCC on main sheet which only connects to 12V sense in on power subsheet?

1. Why is R8 so small? if 2A max this gives you a V range from 0 to 2V - try to use most of the voltage range (for the PIC, 0-5V)
2. I was thinking about this more - we also want current sense for the 12V rail to rocket, and battery charging current
3. Not sure how I missed this one, but the buck converter wants 3 1uF input capacitors. 
image

Layout:
4. C13 and 14 are getting kind of far from the 12V input - probably best to do a 2x2 array of capacitors, if that makes sense?
5. You can move R1 and R2 closer to the buck
If you haven't already, read 9.4 of the buck datasheet, and also [AN-1149](https://www.ti.com/lit/an/snva021c/snva021c.pdf?ts=1738951391516)

Rio6:
I think you mean uH
圖片

I know it's more of a system design issue and the diode is standarized, but is 1A really enough to support then entire rocket indefinitly (unless that's not the requirement)?
圖片

Based on last year, fuse that requires soldering sucks because ones something shorts you have to pull the board out. Might be worth looking into self resetting e-fuse. Most of them have current sense output as well so it can replace the INA.
圖片

Some sort of current limit on 5V might be good too, or at least do some analysis on what happens if 5V shorts. We can also use e-fuse on 5V, most of them supports high-side n-mos switching so it'll give you better efficiency than p-mos as well.
圖片

I'm thinking it might be more useful to get battery current rather than the entire 12V current. This way we'll know whether the battery is charging or discharging. And if every single boards on the 12v charging line has battery current monitor we can even up the charge line current to compensate for rocket's own current draw. Thoughts?
圖片

AshTheEngiqueer (changes requested):
1. No datasheet for U7 ;-;
2. Either source a 455R resistor or adjust your value slightly - I just looked and 455R isnt a value that can be found easily. Also, this is for exactly 2A, which is the same current draw that would kill the buck - maybe set it for 1.9A or even 1.8A? Something like 475R exists and gives a shutdown current of 1.9A
3. You're not using the error output of the efuse?

4. Excellent google sheet, only concern is your estimate of input capacitance? From what I'm understanding, you're saying camera board capacitance is 22uF, estimating all of the boards in total as 3 camera boards worth of capacitance, and then putting a FOS on that? Can you elaborate a bit more on that one?

Layout:
5. Overall nice work on the buck layout/routing, there's a few daisychained traces (R1/R2, for example) and a few things I'd question (is there a reason you're running the GND pad from R2 all the way to the GND fill of C13? Is it necessary to split the GND planes?) but very good work!

Tessa-P:
- no pdf in folder :(

- programming connector has wires running inconsistent lengths under label (this doesn't matter)

- `FLT` should have overline
image

- Note calls out R2 but resistor has since been renamed
image

- add datasheet for can transciever (i hate google)

- Your source impedence on your battery voltage sense line is too high, datasheet says max 10kR
image
image

- fill out title block and put it in your portfolio :eyes:

In general looks good :)

AshTheEngiqueer (changes requested):
Overall looks decent! A few notes before routing:
1. Theres a bunch of stuff that's pretty cramped? I'm looking at the F.silkscreen layer and there's a lot of overlap, partially because stuff is just so close together. Also there's a lot of parts that are slightly offset from each other, which doesnt look very nice.
2. The GND pour for the oscillator should be separated from the general GND pour on both sides, except for one trace.
3. A few things that look like they'll be interesting to route - the far side of the e-fuse (the side farther from the MCU)
4. When you get to it, connectors should line up with the board edge, and the mounting holes should be square

AshTheEngiqueer (changes requested):
1. Fix DRC
image
2. Take a look at fig 2-3 (pg 15) of the datasheet for oscillator layout - you should copy this layout. 
3. center-to-center distance between the mounting holes should be some multiple of 0.5in, also rounded pcb corners please
4. Connectors (screw terminal and gecko connector) should be right on the board edge
5. in general you have some parts that are almost aligned but not quite? the board looks nicer if you can align components so they're all facing the same way (and so silkscreens are also all in one direction)
6. How did you calculate the spacing for the via fence on the oscillator?

StarlightDescender (changes requested):
1. totally missed this, please revert your changes to both RA and charging on the pad, and also tidy up the symbols/footprints (delete stuff you arent using - ex. don't need the L and M SOT-23-THIN footprints)
2. Is there a reason you have CAN TX and RX on non-adjacent pins?
3. Lots of places where you have components really smushed together - this wont be very fun to solder. You have lots of space in other places, maybe there's a way to take advantage of that?
image
image
image
4. Why run the CAN traces through two vias each?
image
5. There's some silkscreens that will get covered by components - L1, U4, and a few others.
6. we cant have parts with thermal pads on both sides? currently we reflow one side, and then we try to reflow the other and then the first side melts and falls off which isnt great
7. thermal pads should probably get thicccc traces so they can effectively dissipate heat. Something like 0.05in or so

Joe-Joe-Joe-Joe:
Spreadsheet is good. I wish every board designer did this
Schematic:
5V
1. Does FLT need an external pullup? Can the PIC internal GPIO pullup source enough current? This is the only reference I could find and it says "no"
image
image
2. schematic note for what this does?
image
3. This is kind of really broad but have you considered transient protection/reverse polarity protection? Most importantly is downstream, we don't want to reverse-voltage the 5V bus
image
main
4. There is a builtin function for DNPs that makes it more obvious
image
12V
5. Pullup for FLT - as above, schematic note for OVCSEL - as above
6. Why do you use 0.01uF for Cin if datasheet recommends 0.1uF?
7. Your schottky diode for the charge line is rated to 1A average fwd current. But your fuse is 5A to the battery (also consider that when charging you are also powering the entire bus. So the minimum current over the diode is (battery max charge current + 5V buck max input draw + 12V (2A since RILM is 475))
8. Why is F1 a chip fuse instead of a replaceable clip fuse (I think there is a standard littlefuse holder we use elsewhere?)
9. L1 is 42mR typ but Webench calls for 19mR
image



PCB:
1. Building on the charge current concerns above - are you exceeding the max current on a single gecko pin for the charge line? 
image
2. standard hole spacing I believe is in increments of 0.5"
image
3. Dangling trace
image
4. I'm not going to check all your power trace sizes. But note that your cross-sectional area across this pour where it constricts is less than the width of the actual trace into the pin. You can almost certainly eke out a bit more space on the top-right of the gecko mounting hole
image
5. Bend osc traces in a "Y" shape rather than 90 deg
image
6. As Ash mentioned, try to spread components more evenly
7. Your edge cut is not straight
image
8. Minor nit, rn your buck gnd plane goes through your pad. you can move C12 and R2 to give yourself a larger path 
image
9. You can move this to the side to match and not overlap ss
image
10. Can move C13 upward to bridge pours directly, then add more GND vias below
image
11. Nit silkscreen labels should be oriented the same way and not overlap vias
image

thats enough for now

StarlightDescender:
Joe raises a good comment re: externall PU resistor on FLT signal lines from both fuses - I see this still isn't addressed?

StarlightDescender:
Still haven't addressed my comment wrt both e-fuses needing to be on the same side?

Pdada1:
> Still haven't addressed my comment wrt both e-fuses needing to be on the same side?

Was thinking we can just use hot air and get around this. Rio is pretty confident it should be ok

StarlightDescender (dismissed):
1. 
image
this is too thin if we're expecting anything close to 5A - 5A with 25C rise has a trace width of 1.6mm, so you should run a thicc trace to here rather than rely on zone fill. Same thing with the 12V output (and ideally we're not using a 25C temp increase)
2. ...what :crying-emoji-dies:
image
image

You should use one of the two central planes as a power plane.
3. In general traces are a bit messy (leaving pads at weird angles), silkscreens are overlapping with each other (or pads)

rexblade21 (changes requested):
image
1. I used 330 ohm resistors for all three colours of LED on MCB and it was very bright already. Doesn't really matter though.
2. larger bulk decoupling capacitor would be nice at the LiPo input, and output of the two efuses. maybe like 2-3 22u ceramics in parallel and a larger electrolytic at LiPo input?
3. Looks like traces from the efuses can still be a bit wider. Also would be nice if it sort of curved outwards asap (i think its called a teardrop?)
image
4. The ~1mm internal traces can only carry ~0.9A at 15 deg rise. I believe the default for internal layers at JLC are 0.5oz copper so consider making traces wider, not routing on internal layers, or paying for 1oz internal copper.
image
image

Joe-Joe-Joe-Joe:
LGTM beyond Jason's comments

Joe-Joe-Joe-Joe (dismissed):
LGTM

rexblade21 (dismissed):
LGTM

image
FYI i don't think you'll be able to find 47u caps in 0805 package with >12V rating. 22u should be fine.

StarlightDescender:
good point, thought i saw some earlier but ig not. i'll update the schematic

StarlightDescender:
shoot i forgot git would do that

rexblade21 (dismissed):

Joe-Joe-Joe-Joe (dismissed):

Pdada1:
image
can probably condense these traces by overallping the resistors
image
also do we need to consider reverse polarity protection for this electrolytic

Pdada1:
image
pls show gain calculation on the schematic

ManavToor (commented):
1. Refer to datasheet 7.3.4. If you tie V- to GND, the op amp wont perform as expected close to 0V.
Screenshot from 2025-05-17 10-05-01
2. Make the connection from oscilator GND to actual GND on the GND plane. Also should move GND plane to layer 2 and power to 3 Also, any reason layers 1 and 2 don't have a GND fill?
image
3. The vias should be much closer to the 5V output of the inductor. I would spread them out throughout the fill.
image
4. Why the via, you can connect it straight down. The two traces should be aprox the same length btw.
Screenshot from 2025-05-17 11-11-14
5. Very nitpick, but a lot of these thick traces can just be made fills
image

Pdada1:
one question I do you think it's a relevant concern that since the efuse is by default open untill the pic disables by sinking, if the pic dies, the efuse will be open by default. Idk if the senario where the pic dies, while battery is above 12V and we are running 12V from gnd is a concern. Idk how a large current movement since the only resistance is RDS of mosfet (100mR) will cause a pull up to lipo voltage or ground side is strong enough to remain pulled down at 12V. In any case is this something rpb needs to handle or will the ground side 12V deal with it?
image

Joe-Joe-Joe-Joe:
Not sure I understand your concern. If you are talking about the charge line, there is a diode to prevent the lipo sending anything back to groundside. Groundside itself will always try to push 1A through the line while it is on (ash is cooking a voltage limit for that today I think). In the case that draw on the "12V" line is <1A, that comes from groundside, and the remainder gets pushed into the lipo. As soon as the draw on that line exceeds 1A, now your liPo is discharging instead. I was concerned about this before but I think with the groundside scheme it is ok

Joe-Joe-Joe-Joe:
> Refer to datasheet 7.3.4. If you tie V- to GND, the op amp wont perform as expected close to 0V.
Good note but is this a real requirement? I feel like we are probably not relying on the current sensing to tell us we aren't drawing anything on the 12V line. And that would require a negative supply, so not worth doing unless we actually need it.

Pdada1:
> Not sure I understand your concern. If you are talking about the charge line, there is a diode to prevent the lipo sending anything back to groundside. Groundside itself will always try to push 1A through the line while it is on (ash is cooking a voltage limit for that today I think). In the case that draw on the "12V" line is <1A, that comes from groundside, and the remainder gets pushed into the lipo. As soon as the draw on that line exceeds 1A, now your liPo is discharging instead. I was concerned about this before but I think with the groundside scheme it is ok
-----------------------------------------------------------
Sorry for being unclear
image
I'm basically thinking about what happens if the pic dies in relation to the orange line. Like in that case aren't we connecting lets say 12.6V of full lipo to the 12V line. Idk what the effects of that is.

Joe-Joe-Joe-Joe:
I think this is actually just a poor choice of power rail name. The rocketCAN "12V" line has always meant "unregulated LiPO voltage". So 12V +- 0.6

Pdada1:
But like in any case if there is a difference in there voltage between the lipo and the whatever voltage groundside 12V is supplying then we will have current flowing in the orange line right. I presume the voltage that the groundside 12V is supplying is not going to be the same as the current lipo voltage

Pdada1:
image
The 400k doesn't seem to be in stock on digikey. Maybe you can shift to 402k. Falling Threshold to 9.81, Rising threshold to 10.848

Pdada1:
Porting over jasons concern from slack: 
1) 
image
Jason wants this current sense resistor moved post fuse 
https://waterloorocketry.slack.com/archives/C07MX0QDS/p1747854024253339?thread_ts=1747853998.586919&cid=C07MX0QDS

Pdada1:
image
No silkscreen for + and - for lipo ring terminal

Joe-Joe-Joe-Joe (approved):
LGTM

StarlightDescender (approved):
reviewing my own commits is kinda cringe but here we are ig

PR #94, opened 2024-11-27
USB Debug Respin

taisirhassan:


AshTheEngiqueer:
1. The boost converter schematic is quite cramped and a bit difficult to read? You have plenty of space in the top row to the left - you can move things that way to give the boost converter more space.
2. Add the link to the boost converter schematic so that we can more easily check your components :)
3. 5V is just directly connected to GND? That can't be right.
4. Use the SOT-23-5-Handsoldering footprint for the boost converter, this will make soldering the chip by hand easier.
5. You'll need to pick a specific inductor and add the part number as a text note, and make sure you download and include the footprint. Most inductors aren't standard sizes, so it's unlikely you can use the 0805 footprint.
6. Add the part number of the diode in text as well :) Don't need to give the voltage drop of the diode, that's not particularly useful.
7. power should point up and gnd points should point down - your 12V power is pointing sideways. Just for aesthetics and ease of understanding :) 
8. At the moment, you're creating a 12V line, but that isn't going anywhere? you'll probably want to replace the 1x4 connector with a 1x5 so you can output the 12V as well.
9. Make sure to run the "annotate schematic" tool to make sure all your components are numbered properly.
10. It's nicer if you change the "TestPoint" text to say what the test point actually is :)
11. VSS on U1 (the PIC) needs to be connected to GND

AshTheEngiqueer:
It would be nice to have the ability to select from either the 5V boost or an external 12V source (like a power supply) - best way to do this is probably a screw terminal for 12V in, and then a 3 pin jumper with the 12V external and 12V boost on the ends and the 12V out in the middle?

PR #95, opened 2024-12-03
Processor 2025 Production Version -  Switched ST IMU from +-32g to +-16g

Armaan-Sengupta:
Production version of what was ordered from JLC PCB, only change is the ST IMU properties being switched from the 32g accelerometer to the 16g accelerometer because thats what JLC had in stock, also apparently 32 was a fake requirement.


Armaan-Sengupta:
Dont even bother aproving this... cant be bothered

PR #96, opened 2024-12-04
Ltt respin

ManavToor:
Respinning LTT to increase range and remove rfd900x

Designed around cc1200 transciever. I basically copied the entire evaluation board cause rf is hard and their filters and matching networks are already made and tested. Did add an external PA which meant breaking away from cc1200 design (bad) so we implemented a switch for switching between receiving and transmitting to keep the cc1200 network untouched. Second switch is for switching between antenna 1 and 2. There is a second PA that we also want to test, hence all the DNM's.


AshTheEngiqueer:
Please remove commits from RA and DAQ CAN support :)

PR #97, opened 2024-12-04
Ltt respin

ManavToor:
Respinning LTT to increase range and remove rfd900x

Designed around cc1200 transciever. I basically copied the entire evaluation board cause rf is hard and their filters and matching networks are already made and tested. Did add an external PA which meant breaking away from cc1200 design (bad) so we instead implemented a switch for switching between receiving and transmitting to keep the cc1200 network untouched. Second switch is for switching between antenna 1 and 2. There is a second PA that we also want to test, hence all the DNM's.


Rio6:
2. is not fixed
6. I think L1 and R8 should share the same footprint
7. one of these 2.2uF should be bigger, like 1nF since the original reference design only had 1 of them. Also 15nH is only 84ohms which is a bit low, we want it >> 50ohm
圖片

Rio6:
I just realized RF_IN on PA is missing M1 and M2 from datrasheet
圖片 圖片

Either remove C42 and add a DC blocking cap here
圖片

Or increase C54 (say to 1nF) so the total capacitance between C42 and C54 is still around 2.2pF

ManavToor:
Updates:
- Decided to ditch evaluation board design and made our own matching network
- Decided to stick with 1 PA instead of being able to switch between 2
- Not using TRX_SW to switch between TX and RX, now using GPIO3 (pg 19 footnote, https://www.ti.com/lit/ug/swru346b/swru346b.pdf?ts=1735522344737)
- Did the layout

Rio6:
1. nit: if we're doing mounting hole with via may as well make it solid 
image
2. nit: they can probably share the same via 
image
3. I'd reduce the via clearance on the ground plane as well so it is not cut off here and some other places 
image
4. nit: maybe remove the two 5V vias on top as well. Or better, move C44, C45 the the bottom-left of the strip image
5. lastly you should fix DRC, check jlc for their size constraints as most errors seems to be from that. Update the net for footprints if the stock one isn't labelled right. And for the clerance ones you can probably just change the DRC setting.

ManavToor:
1. Done
2. nah
3. Added a small fill zone connection between the holes
4. replaced them with just one via directly in between the two caps
5. lowered board constraints to slightly above jlc minimums

Also added a guard ring

Rio6:
For the guard ring, I'd do it like this, excluding the sma connectors and including the transceiver, making it a rectangle. You might have to move the switch or SMA around a bit to make room. We want to put a shield over it, and the SMA wouldn't fit in there. Also rectangle shield is easier to maker. You should add 2 through holes on it for soldering the shield as well (or clips, if we can find one on digikey).

圖片

AshTheEngiqueer:
Sorry I'm late to the party

Schematic:
1. 
image
is this 1A hold or trip current? Also do we want to update CAN connector from datamate to gecko?
2. 
image
I don't think your footprint on the inductor is correct?

Layout:
1. 
image
is there a reason this trace width changes? is it some RF thing? I'm pretty sure you can run the whole trace with the thicker thickness
2. 
image
No via fence on the oscillator?
3. in general there's a lot of daisychaining, but especially with RF stuff routing is giga important? you two know more than I do but maybe worth a look to see if there's anything you can change.
4. Also in the email from JLC there's an extra charge because of the size of some of the vias/holes - either this is something we pay or you can look at increasing the size of vias/holes to minimum 0.2/0.45mm

ManavToor:
1. Hold, trip @ 1.5A. Completely slipped my mind to switch to gecko's, nice catch
2. It matches whats in the datasheet, not sure what you mean

1. clearance errors, the footprint is super tiny and the pads are so close every time I run the trace to the pad DRC looses its mind
2. Added some vias there
3. You have to daisychain for high speed
4. Updated the vias to 0.45mm diameter

AshTheEngiqueer:
This should be merged if it's been ordered

AshTheEngiqueer (approved):

Rio6 (approved):
Board ordered so approving

JasonBrave (approved):

Issue #87, opened 2024-08-02
Remote Arming Issues


BluCodeGH:
- Silkscreens got removed from all the datamate connectors
- Buzzers interfere with mag switches, they need to be moved further away

BluCodeGH:
- PIC-controlled buzzer causes interference with the batt2 current measurement when on

Tessa-P:
Batt current sense should be called altimeter current sense (and numbered respectively)

StarlightDescender:
already fixed in new PR for 2026 but
- swap to geckos for consistency with new boards

PR #99, opened 2025-02-22
Remote Arming V4

ManavToor:
-added 2 resistors in series with BATT and changed fuse footprint


AshTheEngiqueer (changes requested):
- No Lipo charging? Also probably best to update rocketcan connection from datamate to gecko
- R35 through 38 have to have high power ratings and hence you should give a part number, also any reason you're using 2x 2R instead of 1x 4R? please also add datasheet for the specific resistor
- Add datasheet to D2/D3

Tessa-P:
- add datasheet for F1
-> do we have any data on the current drawn by the lipos on fire or on a short?  I worry that 4A is too big.  I'm pretty sure we did 4A last year (that didn't blow on a short) but lipo internal resistance will be better than 9V, so idk.  This might be worth testing before we commit to this fuse?
-> also maybe add the fuse size as a note on the schematic?

- same Q as ash abt R35-38, why not just 1x 4R resistor?  And also why no lipo charging?
-> also just for my own curiosity, what are these resistors for?  

- do we care to address the other issues noted for this board?  mainly the buzzer, but also anything else in the issues page?

ManavToor:
- Updated Datamates to Geckos
- Added Charging
- Changed double 2R to 3R @Tessa-P this is so that current is limited to max 4A (I had originally done 2 2R cause the single resistor was expensive)
- Added datasheets to stuff
- Everything else in #87 will be addressed in layout

StarlightDescender (changes requested):
1. Gecko pinout is wrong, look at altimeter or rocket power for correct pinout
2. Change the 'value' of the 3R resistors to the part number
3. Please remove changes to other projects (ltt, charging on the pad, altimeter)

Is there a way to layout the connectors such that they're on two opposite sides (battery can be on a third side)? I'm thinking about recelec design and having only two opposite sides with connectors makes harnessing/recelec easier to manage

Joe-Joe-Joe-Joe:
1. Your current limit on the batteries is 12V/3R = 4A, but your batt line current sense is sized for 10A. from the design doc it looks like 4A is the correct number?
2. I don't understand how our recovery system works but I also don't understand this label convention. We don't have pyrocutters anymore right? So what do the ematches actually connect to? I would like to be able to wire the altimeters and ematches myself based on the schematic, if it came down to it, and right now I don't think I could
image
3. Your current limiting resistors are rated to 3.5W in free air. Is there any case where you would draw >0.39A continuous from a single battery? I doubt it since the time over which the ematches fire is very small, but just confirming

Hexamech:
image
- Shouldn't the resistance of these be 12mOhms?

StarlightDescender:
- Don't know how i missed this one prior, but 100mA max on the LDO seems pretty tight. This powers one of the buzzers, the PIC, the 3 LEDs, and the CAN transciever. LEDs are 10mA each (using the Vforward and the series resistors), buzzer is 30mA (from datasheet), so that leaves 40mA between the transciever and the PIC? seems sus to me.

Pdada1 (commented):
image
One thing that I think hasn't been though about is mech integration. 
Right now the power resistors are pretty close to the magswitches and looking at thier datasheet they are 20mm high which is bigger than the standoff clearance of the magswitches (0.45in or 11mm) 
We should check to make sure the clearance is suffcient otherwise the resistors will hit the plate on the switches

StarlightDescender:
- via in pad here 
image
- overlapping silkscreens/vias/pads 
image
- in general its nicer to have all the silkscreens facing the same direction
- update board name/version/date?
- daisychaining
image
- there's a fairly thick trace (40mil) between the PMOS and pin 5 of the altimeter connectors thats good for ~3A - does it actually need to be that thick?

PR #104, opened 2025-08-04
change r12 so it actually sets uvlo/ovlo corectly for the 12v line

Lucifersan:
we typoed a resistor real bad


Joe-Joe-Joe-Joe (approved):
LGTM

PR #110, opened 2025-10-12
CanardBoard Full Review

### LINKED ISSUE #125: CanardBoard Review Todos
- [ ] double check all pin assignments across schematic, google sheet, and cubemx file

Schematic sheets to check
- [ ] 32g_imu
- [ ] ad
- [ ] baro
- [ ] flash
- [ ] mag
- [ ] movella
- [ ] power_rails
- [ ] processor
- [ ] pwr_in
- [ ] pwr_mux
- [ ] rocketcan
- [ ] SD_card
- [ ] servos
- [ ] usb
- [ ] top sheet (canardboard)

other
- [ ] Canardboard layout
- [ ] Footprints (pin numbers, sanity check dimensions, etc)
- [ ] Do JLC PCBA stuff

ad breakout
- [ ] AD breakout schematic
- [ ] AD breakout layout

jhuang2006:

The requirements for this project can be found at [LINK](https://docs.google.com/spreadsheets/d/1JLKeAksTf0xKj0QCYFTEEwu3voSvOGnbUyV6GcdOs_8/edit?gid=1738690971#gid=1738690971).

Board calcs (WIP): [LINK](https://docs.google.com/spreadsheets/d/1TvnVFZS-V2CKqBLQlujgA8WVB1qmG7LbKLAjdNWW7EU/edit?gid=0#gid=0).

Design doc (major WIP): [LINK](https://docs.google.com/document/d/1uaVwPAHg6lDtzCRZP6kGtQEOIt_x4iiLFU4eh1LqruA/edit?tab=t.0).

Schematics planning: [LINK](https://docs.google.com/spreadsheets/d/1R60l7a1GPZrk1-Guafgqmg8lstQdjkIb5YvL78OFDvg/edit?gid=0#gid=0).


celery6 (commented):
1. is the TPS's SWEN pin overrated? Why not connect it to mcu as gpio input for extra info
2. are the servo pwms on different timers because pin conflict? Would be nice to use 1 timer with 4 channels

other stuff looks ok from firmware pov including cube file, breakout, and COCKS schematic. Also I agree with joe the CAN names are more readable with underscore

FinnBreu:
grafik
Can you make this prettier?

FinnBreu:
Why does the accelerometer in the compass need two interrupts?

StarlightDescender (changes requested):
schematic
- NRST has an internal weak pullup on it?
- I don't think the BOOT0 jumper is really that helpful, probably more useful to have a small pushbutton for when you want to bootload via usb. 
- why 3A polyfuse on Vbus when Imax is 500mA? 
- Do you need a pullup on flash CS? 
- What's your nominal 12V draw? Also please dont draw from both CHG and +12V.
- best to have specific part numbers for specific capacitors (ex. your Cin on the power rail page, which experiences up to 25V)
- Why do you have separate I2C lines for the baro and 32G imu?

celery6:
> @celery6 is Movella over CAN worth it? If there is one bus just for Movella we could just stick to UART

it's not necessary but I wanted to try using CAN to avoid using the sus parser lib. CAN has better framing.
is this a problem for layout? if it's a problem then we can survive with uart perfectly fine too

rexblade21:
> > @celery6 is Movella over CAN worth it? If there is one bus just for Movella we could just stick to UART
> 
> it's not necessary but I wanted to try using CAN to avoid using the sus parser lib. CAN has better framing. is this a problem for layout? if it's a problem then we can survive with uart perfectly fine too

no its fine. and you want uart still in case we skill issue?

celery6:
@StarlightDescender why would we put the i2c devices on the same bus? That permanently eliminates the possibility of parallel reads. And possibility of different i2c configs for each device (one of them seems to handle faster clk maybe)

FinnBreu:
grafik
Why not quad SPI? And does this config work

FinnBreu:
grafik

Good limit, at 50k the rocket probably breaks from G-forces

FinnBreu:
grafik
Not sure how good the analog signal is here

FinnBreu:
grafik
Some islands have no connections

FinnBreu:
Looks pretty good overall, did not find anything major

celery6:
> @celery6 shuffled stuff around so servo pwm uses TIM1 CH1, CH2N, CH3, CH4. can't use CH2 because it conflicts with QSPI CS, but i don't think theres a difference between CH2N and CH2 for simple pwm output. are you happy with all other pin assignments

@rexblade21 ch2n is fine but would require some config. Can we actually replace qspi CS with a random gpio for CS? I think it's actually better to use software CS pin instead of HAL SPI's "hardware NSS" because it's probably not what we need for flash transactions anyway.

Pdada1:
Schematic looks fine to me

celery6:
@StarlightDescender said to not ground mounting holes now

jhuang2006:
> @StarlightDescender said to not ground mounting holes now

why? doesn't grounding the holes allow the baseplate to act as a shield, separating motors from sensors?

FinnBreu:
Do we really want to board-mount the Movella? It is not required. And the housing is by far the largest component, you could have so much more room for placement and routing if it wasn't there. Also the removed holes wouldn't constrain the MCU placement and routing

jhuang2006:
still working on silkscreen lol. do you want like axes next to each sensor that matches their orientation? and a global board one (like the convention we're using after rotating all the sensor axes to line up)? @FinnBreu

dare-eee-enn:
image
- These should be the INA180A3, the A1 is 20V/V

image
- power to servo kinda gets bottlenecked by these vias
image
- Similar deal here

Silkscreens that indicate which motor is which might also be helpful though I know you're working on silkscreens

Otherwise looks good to me

FinnBreu:
- [ ] Is there a better spot for the barometer, other than between the Lipo leads? I'm uncomfortable with solder access and board stresses affecting the measurements. Heat flow should be fine I think

- [ ] Servo 3 power supply is still bottlenecked by 1.5mm wide passages, both above F6 and U17. Could you add polygon routes in another layer with a bunch of vias? Or rotate the entire L1 group by 90deg

- [ ] L1 is directly next to the Movella, I'm concerned about its fields affecting the magnetometer

- [ ] Servo 1 has the same size bottleneck below D9, can you move the GND via? Maybe move the D9/F4 group up?

- [ ] You'll probably want the AD board symmetric on the main board x axis, but you could move it down to clear space for the power rails / servo 3

- [ ] Is this sufficient? 
grafik

ManavToor (commented):
Board looks good, didn't find any major red flags in the schematic, just a tiny thing:

1. There is already an internal pull up on NRST (fig 18 in datasheet)
Image

The layout looks great too. Is there a reason you don't have a GND pour on layers 3 and 6?

1. Please run DRC, you have a short between your CAN lines and mounting hole
Image

Other than these two things LGTM

jhuang2006:
> image
> Using Vsense of 20mV (20A across 1mR), RIMON of 33k, RSET of 100 gives a voltage of 5.94V which exceeds maximum on pin input. While ig your circuit breaker ILIM is set at 12A, there is still a 1ms blanking interval for which this voltage could be applied to the STM pins. image

thanks for pointing this out! I think we have two realistic options here: pick a R_imon such that 3V3 is not exceeded at the set short circuit current (instant efuse trip) or clamp it with a diode. Given we don't really care about having super high resolution battery current data, I'll just go with the first option.

dare-eee-enn:
image

- I'm curious to what the 0 ohm resistor is used for?
image

- rotating these capacitors might give a better connection to vbat and also kinda shortens a trace
image

- Why are these resistors different values? 


image
- could very well be something on my end but drc is mad at the vbat shorting with like everything

jhuang2006:
> image
> * I'm curious to what the 0 ohm resistor is used for?
setting the short circuit current threshold. 0 ohms just happens to correspond to 20A
> 
> image
> * rotating these capacitors might give a better connection to vbat and also kinda shortens a trace
yep thanks
> 
> image
> * Why are these resistors different values?
red has a lower forward voltage than blue and green.
> 
> image - could very well be something on my end but drc is mad at the vbat shorting with like everything

uhh my drc looks fine... is anyone else having this issue?
image

jhuang2006:
> Not seeing any issues with footprints, maybe make sure whatever leds are being used can handle the current being put through them, green and blue leds on the mcu page have 33mA going through them

how did you check the footprints? did you cross check pin number on the symbol with intended function on datasheet and pad number on the footprint? 

LEDs have much less than 33mA going through them since they have voltage drops of around 3V for blue and green, 2V red

StarlightDescender (dismissed):
power_mux
- no datasheet for RB520S30T1G
- no TVS diodes on inputs to muxed line? VCHG/12V_RKT/VBAT/etc?

sd card
- nit but mismatch is spelt wrong

CAN
- 5A fuse on the charge line is *way* too much, the max output is 1.5A to power all boards + charge main lipo

FinnBreu (dismissed):

FinnBreu (approved):
- Coordinates of AD breakout are correct
- Canard board coordinates are correct
- Movella coordinates are correct
- STM IMU coordiantes are correct

Only two little things:
- [x] STM mag coordinates has swapped the x and y axes. This is not relevant because of how we're using the magnetic field in nav, but fyi
- [x] Can you add "IMU", "Baro", "Mag" in silk next to the ICs/coordinates? For not making mistakes in transforms or conformal coating

This looks really good, great job!

dare-eee-enn (approved):
LGTM, not seeing anything wrong

dare-eee-enn (approved):
computer died on the review page so it exists twice

Pdada1:
image
Any reason we can't fill the B CU with gnd and give all the traces running a reference plane if its not being filled otherwise. 
image
might give you increased shielding so long as the areas are not that thin.

jhuang2006:
> image Any reason we can't fill the B CU with gnd and give all the traces running a reference plane if its not being filled otherwise. image might give you increased shielding so long as the areas are not that thin.

Cuz i'll need to carefully add stitching vias to avoid gnd loops which would further cut up the power planes on top copper.
idk idt its worth the effort lol

Pdada1 (approved):
LGTM

celery6 (approved):
yay

PR #111, opened 2025-10-13
Initial Draft of USB_CAN Debug Schematic

eozleviUW:

The requirements for this project can be found at LINK.


ihsan-sa:
1.  USB data nets don't match in name (DP/DN and D+/D-)

image

2. 12V is  not a common USB-C PD voltage level. I would therefore change the from +12V to something more generic (VCC). On that note, if you're powering this from a laptop port, then the max voltage will probably be 5V. In that case, you don't need a buck converter. But then, if all you need is 5V, then you don't even need the PD stuff bc you can just pull CC1 and CC2 down with resistors (I think 5.1k but don't quote me on that). That's an easy way of getting 5V.

image

PR #112, opened 2025-10-19
GPS with New Connector

Pdada1:
Some schematic changes as follows
-Oscillator changed
-JST connector eliminated
-PIC 18 Pins changed
-Antenova chip footprint changed to QFN

Otherwise mainly layout change for new harwin connector
https://docs.google.com/document/d/1MVI2FVcVLvdXN6-vZZdgDN41M7ggt62CCTXvSoYcLLI/edit?usp=sharing
Rough Design Doc


Rio6:
1. how did you calculate this width? I am getting ~1.3mm assuming JLC is using 60mil substrates for 1.6mm boards
圖片
圖片

2. since there's some space  it might be worth adding DNC pads for E2, E3 pull downs, I remember using them to debug a while back.
圖片

3. I think the pic GND can be connected better than this tiny trace. The return current for UART for example might flow through this pin and you are directing it into your crystal area. What's the reason for having a 3v3 pour?
圖片

4. decrease the pad clearance so the gps ground is better connected
圖片

5. nit: untwists these traces
圖片
圖片
圖片
圖片

Pdada1:
https://www.digikey.ca/en/products/detail/taoglas-limited/EMPCB-SMAFSTJ-B-HT/3522337
Reminder to make sure footprint is compatible with existing edge connector on LTT to reduce BOM

StarlightDescender (changes requested):
schematic:
- CAN can't be on RA. move to RB or RC. (see pg 266/268 of PIC18 datasheet)
pcb:
- 
image

- weve come to the conclusion that, although in theory its best to ground mounting holes, in our case it is just worse, and NPTH mounting holes are best.

StarlightDescender:
what shield connection are you talking about?
according to liz its best to soft ground mounting holes (tie to ground through a small resistor?) or just not bother, and for GPS it doesnt really matter wrt connecting things to chassis ground because it gets mounted on a fibreglass plate anyway.

Pdada1:
> There was an aluminum plate under GPS board for 2024 rocket to shield it from processor, which was grounded with a solder bulb to mounting hole. I don't remember if it was there this year or what the shield plan is for next year
Hopefully with a can clipped that should be good enough.

Pdada1:
Added RF cover clips
https://www.harwin.com/products/S03-15100300R
This is the can which is looking to be used, paired with the clips listed below
https://www.harwin.com/products/S1721-46R

.gitmodules
Rio6: why is gitmodule removed?

StarlightDescender (dismissed):
Good work, a few notes/questions
- please ground mounting holes
- nit but silkscreens are a bit messy, could use some tidying up

otherwise LGTM

Lucifersan (approved):
LGTM :)

StarlightDescender (approved):
lgtm :)

PR #114, opened 2025-11-20
Remote Arming Pull Request

hai-phong-duong:
Remote Arming electrical project. This requires a design review from electrical leads.
We have consulted the electrical standard and relevant documents. 
We ignored errors that relate to the new Harwin Gecko connectors and errors that were deemed OKAY to be ignored by project lead. 

Contributors:
Manav Toor
Andrew Zheng
Hai Phong Duong


ManavToor:
* Please provide a datasheet link for the Fuse F1
* Please change all the Gecko Connectors to an 8-pin connector on the schematic

* It looks like you still have the old gecko footprint, the new one has 8-pins
* You added dimensions and comments on the F.Cu layer, please move these to the User.Drawings layer
* For your mounting holes, please select a footprint that has a plated mounting hole
* One of the goal's of this respin is to make the board smaller, but some of these resistors are way too close to one another, they should be roughly one "ref des" apart. 
image

* This trace could see up to 4A of current, it is way to thin
image

* A single via is not enough for all this current, use more
image

* The three traces can be combined into one larger trace. One of these traces is going to the LDO, it should be thicker since it is a power rail, and should not come from just one via (same reason as last point). The trace going to the INA should be routed separately from the rest.
image

* Once again, traces providing power should be a little bit thicker, there isn't a lot of current going through these guys but I would buff them up a bit
image

* Don't daisy chain traces like this
image

* Why is the trace so thin all of a sudden? It should be way thicker. Also, do not drop a via in the middle of the trace, have it off to the side
image

This is a good start, but there are a few more places where the traces are too thin or components are too close.

Pdada1 (changes requested):
image

 vias too close to pad
image

Via is too close to footprint pad here (is intersecting)
image


image

image
You can reduce the daisy chaining and length of these traces by adding vias here

image

https://waterloorocketry.slack.com/archives/C8VL7QCG0/p1717531280227019?thread_ts=1717530376.571639&cid=C8VL7QCG0
According to slackumentation 2 ematches in parallel draw 2.5A. Based on this 
the thickness of the firing traces at 0.25mm seems too thin
Confirm with @ManavToor 

image
Eliminate Daisychaining
image
image
image

Should put in some vias for stiching between the two GND planes
https://resources.altium.com/p/everything-you-need-know-about-stitching-vias

hai-phong-duong:
@Pdada1 I made all the changes in the lastest push, except the ematch traces.
Last year, the traces for M1+ and M1_PYRO (etc.) were 0.7620mm (which is way thicker than 0.2500mm currently). 
I'm going to confirm with Manav if that's what we're doing this year.

ManavToor:
Pull ups don't need to be thick
image

You have some DRC errors that you need to fix ie
image
image

When switching layers with thick traces, try using multiple vias (2-3)
image

The thinnest points in this fill defeat the purpose of using the copper fill in the first place. Also once again, use more vias on when switching sides (refering to the right hand side)
image


Add some silkscreens: label the connectors for what they do,  add the "MCLR" indicator on the programming header, label the buzzers, and batt inputs (batt1, batt2). Also name the board,, add the waterloo rocketry logo as well as your names (refer to the previous copies of RA).

If you have time, it's more aesthetic to have rounded corners.

hai-phong-duong:
Done. I'm waiting on clarification for the trace thickness of the jumpers.

"Mochi" :P

image
image

Pdada1:
image
Reminder to update your design rules checker constraints to match the electrical standard. 
image
With this added there are a lot of clearance violations and min thickness to look at

StarlightDescender (changes requested):
Sorry it took me so long to get around to this, thought I posted a review earlier but had only written it.
- 
image
still a bunch of daisy chaining on the resistor pullups for the arming FETs, it should be pretty easy to fix.
- Pad 3 of Q4 and Q5 could use a second via on that trace
- silkscreens are a bit messy, some are under components (F1, polarity indicators on the battery terminals), in general we want to have them all facing the same way but that's not super important.
- Also nitpicky but rename IN1.Cu and IN2.Cu to GND and 5V for clarity

Pdada1 (changes requested):
Good work to push this out, just some small changes and then I think it is pretty much good to approve

Nit
image
Mirror this horizontally, makes it easier to follow if input to output flows from left to right

image
Also run the cleanup tracks and vias tool to get rid of stray traces you made during design
image
Change the spoke angle to 90 here to get a better gnd connection
image
same with here
image
daisychaining, easy fix below
image
image
same idea here
image
daisychaining
image
give yourself more clearance here
image
daisychaining
image
image
image
Give yourself a bit more clearance between via and pad on these

Pdada1:
image
maybe a rotated layout like this might work?

Pdada1:
image

Pdada1 (approved):
I don't see any more changes that need to be made. 
LGTM

StarlightDescender (changes requested):
Looks good overall! a few changes
- team wants to switch to 0603 for passive components (resistors capacitors), i know this is a bit annoying but placements will still the same, you'll just need to swap the footprints and then extend the traces a little bit.
- some daisychaining/90 degree traces
image
image
image
- Please label the pyro connectors on both sides, the mag switches need to be on the side that face out so the current connector silkscreens are facing into recelec
- also move the resistors a bit further from the plated holes for the mag switches

dare-eee-enn:
image
image

A bit of daisy chaining

Pdada1:
image
There are currently a lot of DRC errors for the capacitor and resistor footprints which are marked as differing from the library footprints. Is this intentional?

Pdada1:
image
Why not just via directly

Pdada1 (dismissed):

miles-milne:
Board looks pretty good to me. Here's just a couple of my nitpicks. 





Just some more Q3 daisy chaining
image



If possible, have Kelvin routing R14
image



I'd recommend giving C1 an 0805 footprint since they are often of better quality than their 0603 counterparts. 
image



Silkscreen labels for J5 and J8 should not be obstructed by connectors if possible. 
Also "Alt 1 Pyro" having a different text thickness lol 
image



The ground connections to some of the connectors (J5 & J6) are a bit dodgy due to the thermal reliefs and their small pad size. I'd recommend removing the thermal reliefs entirely for a better ground connection. 
image
image

dare-eee-enn:
- PIC datasheet says that R1 should be within 6mm of the MCLR pin (pg 14)
- add a datasheet or calculation for crystal
image

- raw_bat 1 and 2 traces not fat enough as per rocketry electrical standard
image
image
image

- you should put the waterloo rocketry logo on

dare-eee-enn (dismissed):
LGTM

Pdada1:
image
not the most ideal, but via down on one side and shifting of lower taraces to allow for the connection to the fill on the other side. 
MCLR resistor also kept within 6mm

Pdada1:
image
Just looking at the cad and for integration. It would probably be good to swap the buzzer position and this connector, so that this board is only constrained on two sides

StarlightDescender (changes requested):
A few layout notes
- I would change the crystal to an SMD crystal (take a look at power board or gps board for schematic and layout examples) to save space to make routing easier
- R16 and U8 should be moved to resemble something like this
image

- same with U6 and R14
- 6 pin gecko connectors (J5 and J8) too far back from board edge
- We could turn the two solder jumpers into a single jumper to save a bit of space, and there's no scenario I can imagine where we'd want to explicitly pull up one channel but leave the other channel open

ManavToor (commented):
Why is the GND routed like that (U1)?
Image

The copper connection is too thin for the amount of current going through, I would change the thermal connection to solid
Image

Just make the power plane look like this
Image

I know it's tight but the LDO needs to be routed better, try something like this
Image

sky46:
@ManavToor I think you're on the wrong version? Should look like this:
image

My LDO routing is also probably a little dubious though so I'll try and touch it up:
image

ManavToor (changes requested):
Bruh...my bad.

A good start to the board.

Please fix all the DRC warning and errors, especially the broken board outline and overlapping silkscreens.

Also, why are there silkscreens outside of the board?

You might have interference between wires to the 9V and the wires to the strat. Ask @Pdada1 if this might be a problem
Image

In general the routing seems very sloppy, please take a little more care into how you place components and traces:
- Why is this resistor so far away from where it needs to be
Image

- This trace could be straightened out
Image

- Try not to do this, place the via to the side of the trace
Image

- This cap needs to be placed better, also it should be near pin 3 not 5
Image

This is the best way to route a decoupling cap
Image

There's a few things wrong here:
- The LDO output goes from a nice thick trace to the diode to a super thing one headed into one via! Please try to make it look more like the one I showed in the previous comment (the one for the wrong version of the board)
- Why are so many things getting their 5V from that one via, defeats the purpose of a 5V plane
- The routing into that bottom resistor is sloppy
Image

Don't put a via between a diff pair if you can help it, also why is the decoupling capacitor nowhere near the power pin
Image

ManavToor (changes requested):
Looking much better!

- Placing a via in the middle of a trace normally doesn't cause any actual electrical issues but isn't best practice is all, you're placing a hole in the middle of your wire. In higher speed designs it can seriously degrade your signals.

- The decoupling capacitors are looking much better but you don't need to place a keep out zone around them
ie This can be routed as follows:
Image
Image
Same thing here and a few other places, that keep out zone can be removed
Image

- Could you remove the squiggle
Image

- LDO routing is done well

Pdada1:
image
From a usabilty perspective, it would be nice if the position of the battery header and the magswitch were swapped so that only two side of the board are required to be open for wires to run in. 
image
can probably use pin 15 for this instead.

ManavToor:
You don't need to length tune crystals, not sure why it's like that on other boards. The traces should be short (and ideally symmetrical) as possible for protecting from interference. But none of this even matters at 12mhz so do whatever you'd like tbh.

Pdada1:
image
The M1+ trace is part of the main firing circuitry, so it should be as thick as the m1 pyro. 
image
Nit, move the via a bit and make the traces straight. 
image
Not a super big deal, but I think you can reduce the length of the first trace to the fuse if you use the empty space you have next to the connector. 

image
image
@StarlightDescender correct me if I'm wrong but the Strato TX requires a uart module so it should be connected to a pin in B or C port set. 

Also @Lucifersan are there signals for which having testpoints for would be useful. 

Otherwise seems good to me

Pdada1 (commented):
See above

StarlightDescender (changes requested):
- recieving from strato needs a uart peripheral so that needs to connect to a B or C pin
- put testpoints on UART_RX and the output of both op-amps (op amp outputs are nice but not strictly necessary), these can be small pads (circle or square 2mm)
- 3d model for D2 is not lining up with the pads
- 

Image
ground and batt traces from the LDO don't need to be particularly large, max possible current is 500mA

more to come

sky46:
Made all of Pranav's and Ash's requested changes, except finding 4 different buzzer options and adding to schematic comments (will do tomorrow). @StarlightDescender Is the new routing for ground and batt for the LDO okay?

dare-eee-enn:
image

- I feel like we shouldn't daisy chain these decoupling capacitors so the board doesnt completely die if the pad rips off or something
- Also capacitor is really close to the can transceiver so that might want to be moved away a little  

image

- Same thing here for the current sense

image

- again here

image

- move via out of the way and move the decoupling cap close to the mclr pin

image

- Might want to increase number of spokes or widen the spokes to ground pin of battery

jhuang2006:
@StarlightDescender mounting holes shouldn't be grounded anymore right?
image

maybe we can find a mosfet with lower Rds and lower gate voltage (like 4.5V for min Rds). Though this will work just fine lol.
image

you want a 2x 100nF caps, one for VDD and VIO in addition to the 10uF bulk (not strictly necessary). 10uF should probably also be 0805 for derating reasons.
image

Why is R24 15k? can it not be 10k
image

thermal reliefs look kinda small
image

I may be missing something here, but it seems theres only 1 battery, so is D2 necessary? Did we historically have 2 batteries which is why the diode exists?
Also fuse the battery pls
image

jhuang2006:
100mA enough?
image

PR #116, opened 2025-12-29
Ltt respin

ManavToor:

Drop PIC to 3v3
Connect extra GPIO to PIC
SPI test points
Fix PA matching

still haven't figured out if (and how) we want to remove PA switch


Pdada1:
yay, LTT respin 2 electric boogaloo

StarlightDescender (changes requested):
schematic
- U3 Pin 12 should be connected to VCC2 through external resistor as per datasheet?
- why are pullups on enable lines for U3 different sizes?
- U5 datasheet missing
- why is there a power divider both integrated on here but also existing as a separate board? what's going on, my understanding is that LTT was intended to have two antennas and that the one recieving/transmitting a stronger signal would be connected to LNA_P/LNA_N?
- is the 'todo: find an SPST switch' remaining incomplete for this cycle?

layout
- According to the review from guerillaRF, C8 should be much further from the switch?
- Y1 seems to be a different crystal from what the team standard is based on the 3D model? also is placed on the back of the board, unsure if recelec works with this.

PR #117, opened 2026-01-06
Camern't board respin

dare-eee-enn:
Camera Board connector changed, pcb layout changed


JasonBrave (changes requested):
1. Put page number reference where you get "it has internal pullup"

Image

2. Remove Camera connection and related circuit,

JasonBrave (changes requested):
Additionally:

1. Add a `MX25L25645GM2I-08G` flash, refer to Canard board schematics
2. Use a full size programming connector that match the pinout of STLink V3 Mini-E
3. Since we cannot read flash directly off-board, we should keep the USB connector

JasonBrave (dismissed):
Schematics issue: 
1. flash chip do not have correct footprint
2. Why there's two flash io0

Image

3. Normal SPI controller do not work with SPI devices with 4 data pins, a QuadSPI peripheral need to be used, please use the following pinout
image

StarlightDescender (dismissed):
Image
unconnected 2V8 here

you should length match usb D+/D-

after discussion with pranav, i think it makes sense to supply both 12V and 5V as inputs to the buck converter - easiest way to do this is to add a third diode in parallel, from 12V to input of buck converter.

Also should swap the logic voltage to 3V3 and swap out the CAN controller for a 3V3 controller, so you don't need a second buck/ldo.

Pdada1:
image
can give yourself a better connection to 3V3 plane here

Pdada1:
the review button isn't popping up but see comments

PR #118, opened 2026-01-10
Old ltt fixes

StarlightDescender:
Updates to old LTT:

schematic only so far
- change connector to new 4+4 pin gecko
- shrink passives to 0603
- change buck converter bulk decoupling capacitors to 1206 22uF tantalum-polymer capacitors
- change crystal to SMD


dare-eee-enn:
image
Missing capacitor on VIO and VDD capacitor is supposed to be 0.1uF

PR #123, opened 2026-01-26
Usb debug improvements

StarlightDescender:
- change to USB-C receptacle
- add test points
- add mounting holes
- change to a locking connector


dare-eee-enn:
image
What is this 5V trace doing? If you're trying to have better 5V connection to the right side of the board there are missing vias

image
is there a reason why the CAN traces going into U3 are going all the way left before via down to the bottom layer?

image
Your 5V trace also turns the ground plane into islands daisy chained through R6 and J4

J4 silkscreen needs to be moved


image
random via

Pdada1:
image
Datasheet for molex connector links to old TE connector. 
image
datasheet link broken
image
missing datasheet

Pdada1:
image
image
image
image
image
image
New layout with better plane connection idea
see if it makes sense to implement

Pdada1 (changes requested):
See above

PR #130, opened 2026-04-05
Change one resistor value for AD breakout

jhuang2006:

The requirements for this project can be found at LINK.


jhuang2006:
I changed one resistor value on the AD breakout board to correctly size the gyro's cutoff freq (previously ~60Hz, now ~400Hz). @FinnBreu is this fine?

image

FinnBreu (changes requested):
> I changed one resistor value on the AD breakout board to correctly size the gyro's cutoff freq (previously ~60Hz, now ~400Hz). @FinnBreu is this fine?
> 
> image
> 

Depends, is the ADC sampling with >1kHz and adding its own lowpass?
If not, then the cutoff frequency should be  <200Hz (e.g. 120Hz) because the Nav rate is 400-500Hz and we want to keep the bandwidth well below the Nyquist frequency

jhuang2006:
> > I changed one resistor value on the AD breakout board to correctly size the gyro's cutoff freq (previously ~60Hz, now ~400Hz). @FinnBreu is this fine?
> > image
> 
> Depends, is the ADC sampling with >1kHz and adding its own lowpass? If not, then the cutoff frequency should be <200Hz (e.g. 120Hz) because the Nav rate is 400-500Hz and we want to keep the bandwidth well below the Nyquist frequency

image
I think we're ADC sampling at 1kHz which means its internal filter is 483.8. if i'm understanding you correctly, this is not sufficient so I should use a 6.8uF cap which gives the gyro a cutoff of 130Hz? or 5.6uF for ~160Hz

FinnBreu:
> I think we're ADC sampling at 1kHz which means its internal filter is 483.8. if i'm understanding you correctly, this is not sufficient so I should use a 6.8uF cap which gives the gyro a cutoff of 130Hz? or 5.6uF for ~160Hz

Yes. The filter chain has to be tuned for the Nav rate, let's say that is 400Hz worst case.
It must be guaranteed that the system bandwidth is lower than half the Nav rate, ADC filter at 483.8Hz is not low enough so it requires another filter. Simplest way is as you're doing now, cap range is about right. 130Hz is a bit more sluggish but more robust if Nav rate drops to 300Hz

jhuang2006:
ok changed to 130Hz for now we can change the cap later if needed

tristan-yk (approved):
lgtm (what is nyquist on bro)

dare-eee-enn (approved):
lgtm

FinnBreu (approved):
Signals and Systems is gonna be fun for y'all

